PR #21, opened 2019-03-29
Update simple-handler.go

e-ht:


bradleypeabody:
Whoops, good catch.  Thanks.

PR #27, opened 2019-03-31
Added instantiateStreaming polyfill to support Safari

ptxmac:
Safari (both macOS and iOS) have no support for `WebAssemly.instantiateStreaming` (see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiateStreaming)).

This patch polyfills the missing function as suggested by the main [go wasm wiki](https://github.com/golang/go/wiki/WebAssembly) ([code snippet](https://github.com/golang/go/blob/b2fcfc1a50fbd46556f7075f7f1fbf600b5c9e5d/misc/wasm/wasm_exec.html#L17-L22))

### DIFF
--- a/simplehttp/simple-handler.go
+++ b/simplehttp/simple-handler.go
@@ -402,7 +402,13 @@ var DefaultPageTemplateSource = `<!doctype html>
 {{end}}
 </div>
 <script>
-var wasmSupported = (typeof WebAssembly === "object") && (typeof WebAssembly.instantiateStreaming === "function");
+var wasmSupported = (typeof WebAssembly === "object");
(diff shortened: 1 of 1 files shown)


bradleypeabody:
Cool, and thanks for the PR.  I moved the polyfill inside the wasmSupported block so it doesn't unnecessarily error in the !wasmSupported case but otherwise great and it's merged. I'm going to update the vugu.org site with this shortly and try to load it on my iPhone; fingers crossed...

ptxmac:
> I moved the polyfill inside the wasmSupported block so it doesn't unnecessarily error in the !wasmSupported case

good catch!

Issue #9, opened 2019-03-29
vugufmt - make formatter for .vugu files


bradleypeabody:
Can we make the HTML all nice and pretty, do some basic indentation on the CSS, and then run the script tag through gofmt?  We could do this during code conversion, could provide gofmt-like functionality overall and help a lot with the dev experience. VSCode and other editors are going to want to call it too.

erinpentecost:
I'm going to give this one a shot.

erinpentecost:
Here's some WIP: [https://github.com/erinpentecost/vugufmt](https://github.com/erinpentecost/vugufmt)

bradleypeabody:
@erinpentecost  Very cool!

A few quick notes:

* Ideally, I think this should be available as a struct+method call, as well as a command line tool.  As an example of this, vugugen (https://github.com/vugu/vugu/blob/master/vugugen/vugugen.go) is a command line tool that performs code generation, but it is just a thin wrapper around ParserGoPkg (https://github.com/vugu/vugu/blob/master/parser-go-pkg.go).  It's really useful to be able to call it in-process or from the command line depending on the use case.

* I like that you have HTML and CSS formatting in there - I think that would be really useful as it's very common to be pasting around things in HTML files and have the indentation get messed up.  There will be some interesting edge cases to sort out but that will be all part of the fun (for example, if you forget to close a tag, if the HTML parser you're using closes it automatically, should it do this?  Would that be expected behavior?  And it should it be optional?)

* Bear in mind that the .vugu file format is not set in stone.  One thing that might be added soon is a top level `<script type="application/javascript">` tag for JS to be included on the page.  Probably not a huge deal to just add later to vugufmt, but thought I'd mention it.

Definitely let me know if you have questions or want to discuss the design, etc. further.

erinpentecost:
I finished a big rewrite. The vugufmt package exposes this function as its API:
```go
// RunFmt runs vugufmt on input, and sends a pretty version of
// if out to output. If there is an error, throw away output!
func RunFmt(input io.Reader, output io.Writer, simplify bool) error 
```
There's a runnable main package in there, too, which mimics most of the arguments of gofmt.

The HTTP parser assumes everything in a vugu file is in the <body> portion of an HTML file. If you add stuff in a <head>, it will currently be thrown away. That's probably not what you want. If you have malformed HTML that the parser can't figure out, it just won't attempt to gofmt the file.

Let me know what you think, and if there are specific changes you want to see.

[...]

PR #31, opened 2019-04-04
Vugufmt

erinpentecost:
This is a port from erinpentecost/vugufmt to bring the code into the main vugu repo.
It brings in a lot of code from Go's html package, which has been modified and hidden under /vugufmt/internal.


erinpentecost:
#31

erinpentecost:
#9

bradleypeabody:
Thanks for all the work you've put in on this.  One quick note is can you move `github.com/vugu/vugu/vugufmt/internal` to `github.com/vugu/vugu/internal`? (this way the top level vugu code generator can get at your htmlx package and can use its line number functionality)

And otherwise I'm having a bit of trouble getting it to work.  Hopefully it's something simple.

Here's what I'm trying:

I create [demo-site-path]/demo.vugu as:
```html
<div>
    <h1>This is a test</h1>
        <p>Some text here.</p>
</div>

<style>
.test { background: green; }
</style>

<script type="application/x-go">
func Blah() int { return 0 }
</script>
```

With your PR checked out I `go install ./cmd/vugufmt`, it creates the executable and it's in my path. (I checked to make sure it's the latest, based on the code from this PR - commit 04a5a3e).

Then from [demo-site-path] I do: `vugufmt demo.vugu`.  It outputs the same file to stdout with no changes.   Trying with `-d` gives no output, and with `-w` does not appear to change the file at all.

Any ideas?

bradleypeabody:
Okay sorry, my bad - I was looking at that func and expecting it to be broken into three lines.  It is in fact working on the Go code.

Does it do any indentation adjustment on the HTML perchance?

erinpentecost:
Sure thing, I should have some time today to move stuff around.

Unfortunately, this version does not format HTML. Building up the DOM tree and then rendering it back out just wasn't reliable. This version does do some simple error checking on the HTML though and reports where in the file the error is. I figured HTML formatting could be a future change.

bradleypeabody:
Okay awesome, makes sense.

I will try to spend some time this evening reviewing the code in more detail and will look out for an update to the PR with that internal folder change.  Should be no problem to merge it into master very soon.

erinpentecost:
The format function will return an unedited version of the input if there was an error. The second return value will be non-nil if there's an error, and it includes location fields in the source input as well as an error message describing the issue. This makes it so we won't accidentally blow away the file if there's an error by placing the error info in the same return value.

bradleypeabody:
I'm merging this in now.  Great work!

PR #34, opened 2019-04-07
vugufmt HTML indentation

erinpentecost:
This change forces correct indentation *between* tags if there is a line break between them. Indentation inside a tag is not touched (yet).


erinpentecost:
I elected to use tabs, since gofmt uses tabs. Hope that doesn't bother folks.

bradleypeabody:
Great, looks good to me.  Agreed on tabs - I was originally a space person myself but gofmt made me give up the holy war long ago.  Merging.

Issue #10, opened 2019-03-29
Run goimports to help with missing imports in .vugu files


bradleypeabody:
See how feasible this is.  Could help make dev experience a bit smoother.  Might make sense to do as part of #9 (vugufmt)

erinpentecost:
Since goimports formats code, this is a simple swap in vugufmt. I'll try it out.

bradleypeabody:
I'm going to go ahead and close this.  Seems like vugufmt is a better place to be doing this anyway.  The vugu code generator now deduplicates imports which handles some of the problem.  And since vugufmt does goimports, I think we're set on this point.

PR #36, opened 2019-04-07
Goimports 

erinpentecost:
Added tests for goimports.
Added -i flag to vugufmt to handle goimports. Should this be a different executable altogether (vuguimports)?

#10


bradleypeabody:
This looks like it bombs in Travis-CI because goimports is not available in the environment. https://api.travis-ci.org/v3/job/516940129/log.txt

Any ideas on how to fix?  If not feasible I would be okay with just making the goimports test require a command line flag for now (so it doesn't run on Travis-CI but can still be invoked manually from the command line).

erinpentecost:
I can make a makefile that handles gofmt/goimports installation and then runs tests, and then call that inside the TravesCI configuration. Sound good?

bradleypeabody:
I'm not a big fan of makefiles, but could be fine if it's specifically for Travis-CI.  But would it not work to just do (and I'm only reading from the [docs here](https://docs.travis-ci.com/user/installing-dependencies/), I have no experience with this, but seems like it would work):

In .travis.yml:
```yaml
before_install:
  - go get golang.org/x/tools/cmd/goimports && go install golang.org/x/tools/cmd/goimports
```

I'm a big fan of reducing tool dependencies wherever possible.  But again, for Travis-CI it doesn't really matter so I don't want to be too picky.  I'm mostly concerned if it affects the experience for other developers using (or to a lesser degree, hacking on) the vugu library; which I don't think this does.

erinpentecost:
Your change to the TravisCI config looks perfect.

bradleypeabody:
Cool, although I'm realizing now I might need to do something to add it to the path.   I'll give it a shot.

bradleypeabody:
Yay, it's working.

PR #45, opened 2019-05-13
for evaluation by brad peabody

iansmith:
Don't know if you want the dispatcher... but here is the one I did over the weekend.


go.mod
```diff
@@ -1,7 +1,21 @@
...
+	github.com/microcosm-cc/bluemonday v1.0.2 // indirect
+	github.com/pkg/errors v0.8.1 // indirect
+	github.com/sergi/go-diff v1.0.0 // indirect
+	github.com/shurcooL/github_flavored_markdown v0.0.0-20181002035957-2122de532470 // indirect
+	github.com/shurcooL/highlight_diff v0.0.0-20181222201841-111da2e7d480 // indirect
+	github.com/shurcooL/highlight_go v0.0.0-20181215221002-9d8641ddf2e1 // indirect
+	github.com/shurcooL/octicon v0.0.0-20181222203144-9ff1a4cf27f4 // indirect
+	github.com/shurcooL/sanitized_anchor_name v1.0.0 // indirect
+	github.com/sourcegraph/annotate v0.0.0-20160123013949-f4cad6c6324d // indirect
+	github.com/sourcegraph/syntaxhighlight v0.0.0-20170531221838-bd320f5d308e // indirect
```
egonelbre: Note that this adds a lot of dependencies to the package.

iansmith:
I've broken the tests by creating an import cycle.  It's going to be a bit tricky to fix, but I'm going to look at it in the next day or two.

@egonelbre -- I think those are dependencies caused by my color logging. I don't think there is a way to do conditional deps so I guess if people really object I can take out the color stuff.

bradleypeabody:
Thanks for this @iansmith . I'll need to give it some more thought - I'm not sure if it makes sense to have something like Velvet templates integrated into simplehttp, as opposed to something that gets added on separately.  There are certainly a number of possible template libraries out there that could be used.

The other factor is that some of the code that you're touching in ParserGoPkgOpts is probably going to have conflicts with things I'm doing in my refactor, which I hope to have pushed soon - I'll be working on it a lot this weekend.

But in any case I appreciate your taking the time to post the code and I think we'll just need to see how things pan out.

PR #48, opened 2019-06-01
first cut at some lifecycle methods; no change for people not using them

iansmith:
This PR should be a big nothing burger for anybody who doesn't need/use the features.

It does two things:

* Adds a parameter to vugugen that lets you set the name of the component's class. It defaults to "root" so this change means nothing if you don't set the flag.

* Adds some support for lifecycle type events at the component level.  (Should these be on <ComponentType>Data?)

These work by asking you to implement an interface and have your component return that implemented interface via the GetXXX() methods.  If you don't implement the interface, we generate the code in your <component>.go such that the GetXXX() methods return nil and do nothing.

There are two true "lifecycle" interfaces, Starter and Ender.  Not sure if these should be expanded or be more fine-grained, etc.  I don't really have strong feelings about these.

One of the others is CapabilityChecker that called "early" so you can redirect the user to another page if the capability you need is not present in their browser.  For now it tests for html5, webgl, and web workers.  You can return a string that is your "browser isn't good enough" page.

The other is a way to communicate that you have received background data and are ready to update the UI.  You do this by pushing a value into a channel (an interface{}) and when the main event/redraw loop sees it, it passes it back to you (via Update()) and expects to handle it "quickly" (synchronously).

The capability checker has a dependency on some JS functions.  Is there a way to be sure that the js functions are in place before we start executing? I put them in `wasm_exec.js`, but that seems hacky.

```

function vugu_detectHtml5() {
                if  (typeof document.createElement('canvas').getContext === "function") {
                        return true; //extra if because worried about return val
                }
                return false;
}

function vugu_detectWebGL() {
        var canvas = document.createElement("canvas");
        var gl = canvas.getContext("webgl")
                || canvas.getContext("experimental-webgl");
        if (gl && gl instanceof WebGLRenderingContext){
                return true; //extra if because worried about return val
        }
        return false;
}

function vugu_detectWebWorker() {
        if (window.Worker) { //extra if because worried about return val
                return true
        }
        return false;
}
```


bradleypeabody:
Thanks for this, Ian.

So my main concern with this is I think we're both doing changes that will conflict very badly. 

My refactor has unfortunately been going very slowly, but some of the key changes are for example the ComponentType interface will go away and there will not be two types for each component but just one.  This is a big change and touches all kinds of things. There are a few other ones like this.  (Some of this work is on component-refactor branch but it's still very very incomplete.)

I'm not sure what the best way to proceed is.  The motivation for what you're doing is sound, but I feel like merging this in right now is not going to be overall productive for the project if that makes sense.

Re the lifecycle method interfaces - I think should be more fine-grained.  One of the not-so-obvious goals is to be able to implement components in code-only where necessary (no .vugu file, just a Go struct).  In this case, vugugen and friends wont be there to add missing methods and so we'd want to be able to add a new lifecycle method without breaking existing components that don't have that method defined.

Let me know what you think on this.  The last thing I want to do is just crap all over the work you've done here, and I'm trying to balance things and see what the best course of action is.  I also hope to be able to kick it in to high gear on my own development on this and complete some of the key in-progress work I'm on, which I think will help set the pace for the other necessary changes.

iansmith:
Absolutely not crapping all over.  I understand the cocerns about a big merge conflict.

I don't really have any attachment to this stuff, other than I need a clean way to do asynchrous I/O and be sure that the person who is handling the data can absolutely not pay attention to locking or anything else so you can just handle the data.

All I would ask is that as you get your new refactor near completion, lets try to add some of this stuff into *it* so I'll be able to transition pretty easily from this (admitedly primitive) stuff to something better quickly.

bradleypeabody:
> Absolutely not crapping all over. I understand the cocerns about a big merge conflict.

Okay cool.

> I don't really have any attachment to this stuff, other than I need a clean way to do asynchrous I/O and be sure that the person who is handling the data can absolutely not pay attention to locking or anything else so you can just handle the data.

The mechanism I had in mind for this was that you could call DOMEvent.EventEnv() which gives an EventEnv, on which you can call Lock(), apply whatever data changes (for example that came back from an HTTP response made in another goroutine) and UnlockRender() which causes the render loop to run again and update the UI with the new data.  While I do realize this idea is different from what is usually done in existing JS UI's, I think works well as idiomatic Go code and is not difficult to implement.  Maybe we can chat about this and the pros and cons in more detail on Slack.

> All I would ask is that as you get your new refactor near completion, lets try to add some of this stuff into it so I'll be able to transition pretty easily from this (admitedly primitive) stuff to something better quickly.

Yes, absolutely!

bradleypeabody:
I'm going to go ahead and close this now as things have changed a lot in the project since.  Let me know if you want to discuss anything on it further and thanks again for sharing your work on this.

PR #60, opened 2019-07-30
Goodbye htmlx

erinpentecost:
This change drops the htmlx package in favor of inserting a line-counting reader middleware on top of the normal html tokenizer.


bradleypeabody:
Okay awesome!  I'm still using htmlx in another branch with some stuff I'm going, so I don't know that we're quite ready to remove that package entirely (I'll take this into account but will need some time to refactor), but I'm totally good switching vugufmt over to use your line-counter.  If you can give me a PR with just that I can merge np.

bradleypeabody:
@erinpentecost question on this - have you given thought to if/how byteline can be made to work with the higher level `html.Parse` interface?

I see that you're using the `html.NewTokenizer` approach to read a token at a time for bytes.  It makes sense that this could work well reading one token at a time.

However for general .vugu file parsing, I need the higher level `html.Parse` interface.

I have an idea - please tell me what you think of this. What if we:

* Fork/copy `golang.org/x/net/html` to `github.com/vugu/html`

* Add a [ParseOption](https://godoc.org/golang.org/x/net/html#ParseOption) which allows one to specify a callback function (visitor pattern) so that when you call html.ParseWithOptions with this option specified, you will get a callback for each token read.  Something like:

```go
html.ParseWithOptions(r, html.ParseOptionVisitFunc(func(token *html.Token, node *html.Node) {
    // called from inside the parser each time a node is assembled from one or more tokens,
    // node is the node in question and token is the most recent token read
}))
```

* The work would need to be done of going through the parser and finding where to call this visitor function (actually could be multiple functions, no reason to limit it to one).

* We then put in functionality similar to that of your byteline/reader.go (I think for this use case it can be simplified - we wouldn't need the writer) - so there is a simple line counter available in the package. 

* Then, if one wants to examine line number info while using html.Parse functionality, this can be done by just asking the line counter what line/position it is inside that callback function (shown above).

* With this idea working and the needed tools in place, we could also see if there is some way to make it more convenient for general use.  This could mean adding fields like "Line" and "Col" to html.Node, or it could mean making another ParseOption line ParseOptionLineCount that also automatically keeps a map of *html.Node to line+col and a way to get at it.

* You can then either update your [issue with Go project on line numbers](https://github.com/golang/go/issues/31312) or make a new one to discuss the idea of getting this pulled back into `golang.org/x/net/html` - either way we'll have something that works well for Vugu.  And the added benefit at working code to discuss the upstream merge.

Let me know if that makes sense and what you think of it.

bradleypeabody:
I'm going to go ahead and close this as the code base has evolved quite a bit since this time.  Let me know if you want to pick this back up.

PR #83, opened 2019-12-03
Sync html, head and body elements

### LINKED ISSUE #81: html tag attributes are not sync
**Describe the bug**
If I add attributes to my `<html>` tag inside a `root.vugu`, there are absent from the rendered version. I didn't test with other tags like `<body>` but it's maybe impacted too.

**Software Versions**
Vugu version: c4c2de3e9ddf5cfa914d42e7205ade62b4369e74
Go version: 1.13.4
Browser and version: Chromium 78.0.3904.108 (Official Build) Arch Linux (64-bit)

**To Reproduce**
root.vugu that starts with `<html class="whatever">`

**Expected behavior**
The attributes from html tag should be copied from the root.vugu to the rendered version.

tbruyelle:
Fix #81

This a first and naive attempt to sync html tags attributes from root.vugu to the final rendered result. I'm not very familiar with the code base but I think the change should be located in the `vuguRender` js function.

At first I tried to use the existing opcode `opcodeSetAttrStr` but it appears it's really tied to the mount point, and I wasn't able to use it with the html tag. Thus I created a new opcode `opcodeSetAttrHTML specific for the html tag. I'm not very happy with it so tell me if there's a better way.

I didn't expect such optimization for the rendering phase, but I assume this is mainly for performance reason.

I added comments where I'm not very confident.

I also have to fix a small bug in `domrender/renderer-js-script-maker.go`, the generated package name was wrong, probably because of a relocation.


bradleypeabody:
@tbruyelle Thanks for this PR - it's a good start.

> I think the change should be located in the `vuguRender` js function.

Yes, exactly so.

> I didn't expect such optimization for the rendering phase, but I assume this is mainly for performance reason.

Yes - syscall/js makes it easy to get references to DOM elements in Go but it comes at a much higher cost than might be expected (each element is provisioned for both in Go and JS, there is no garbage collection mechanism, and the calls back and forth any time you want data have enough overhead to be quite problematic at scale).  The approach Vugu is using is very explicit about the communication between Go and JS and is an important part of being able to make the library scale to support large applications (and perform well while doing so).

> I also have to fix a small bug in domrender/renderer-js-script-maker.go, the generated package name was wrong, probably because of a relocation.

Thanks, yeah this was refactored recently.

> tried to use the existing opcode `opcodeSetAttrStr`

I think this is going to be the right approach, but it needs a bit more than just this one call.  What is needed is an opcode to "select the HTML element" (put a reference to the `<html>` tag in `state.el`) and then you should be able use the same pattern of calling `writeSetAttrStr` for each attribute that should be there and `writeRemoveOtherAttrs` to remove any others (which emits the corresponding opcodes/instructions).

I believe the following approach will work:

* **Create an opcode called `opcodeSelectHTML`.**  I believe this only needs to set `state.el` and should set `state.nextElMove` to null just be safe.
* **Write a corresponding `writeSelectHTML`** method on `instructionList` (renderer-js-instructions.go)
* **Change the `JSRenderer.visitHTML` method** you created - structurally if it is not going to recurse into it's children, then it should not start with "visit", so maybe just rename it to 'syncHTML" - and have it call writeSelectHTML and then follow the pattern from the top of `visitSyncElementEtc` - it would be smart to extract this block of code out into another function, but I see no reason not to support attributes, properties and events, which would be this code block (borrowed from visitSyncElementEtc):

```go
	for _, a := range n.Attr {
		err := r.instructionList.writeSetAttrStr(a.Key, a.Val)
		if err != nil {
			return err
		}
	}

	err := r.instructionList.writeRemoveOtherAttrs()
	if err != nil {
		return err
	}

	// do any JS properties
	for _, p := range n.Prop {
		err := r.instructionList.writeSetProperty(p.Key, []byte(p.JSONVal))
		if err != nil {
			return err
		}
	}

	if len(n.DOMEventHandlerSpecList) > 0 {

		// store in domHandlerMap
		state.domHandlerMap[string(positionID)] = n.DOMEventHandlerSpecList

		for _, hs := range n.DOMEventHandlerSpecList {
			err := r.instructionList.writeSetEventListener(positionID, hs.EventType, hs.Capture, hs.Passive)
			if err != nil {
				return err
			}
		}
	}
	// always write the remove for event listeners so any previous ones are taken away
	err = r.instructionList.writeRemoveOtherEventListeners(positionID)
	if err != nil {
		return err
	}
```

Note that we don't want the stuff after that which deals with innerHTML and child nodes - those will not be supported on the HTML node.  For `positionID` use a static value, probably just "html" - it just needs to be unique and if it contains no numbers it is guaranteed to be.

And ideally this same thing would be done for `<head>` also, but if you do `<html>` it should be easy to duplicate.

Give that approach a try and let's see how that goes.

tbruyelle:
@bradleypeabody Thanks for the detailed feedback, I will update the PR.

About the head element, I'll try to see if I can integrate it smoothly in this PR. On the other hand I'll address it on a second PR.

Stay tuned :)

tbruyelle:
@bradleypeabody Ok thanks to your detailed comments, I got something working for both `<html>` and `<head>` elements.

I added 2 new opcodes, one to select the html tag and one for the head tag, maybe we could merge them into a single one, wdyt ? I saw that commented opcode `opcodeSelectRef`, we could use it for that.

 I realize also we may apply the same thing for the `<body>` element, isn't it ?

bradleypeabody:
Awesome and thanks!  A few points/notes:

* I agree on merging these two calls - seems like we need `opcodeSelectQuery` and you just pass in the argument to document.querySelector
* And yes the same thing can then be done with the `<body>` tag.

`syncElement` func looks good - much better to not duplicate all that, so thanks for that.

With those changes, we should be good to merge this.

tbruyelle:
@bradleypeabody I have added 2 commits:
- 33425db merge the 2 calls we agreed on. I'm still not very confident about the value to pass in `checkLenAndFlush`, I looked at the other methods and I wasn't able to determine a common rule.
- a172f26 sync the body attributes. Like I did for html and head, I set `positionID` to `[]byte("body")`, but I'm not 100% sure it's fine.

Thanks in advance for the review!

bradleypeabody:
@tbruyelle Great - one last thing - thanks for asking about checkLenAndFlush, the value passed here needs match the number of bytes emitted based on the instruction.  In this case the opcode is 1 byte and the string is 4 bytes plus the bytes of the string itself.  So what you want is:

```go
err := il.checkLenAndFlush(5 + len(selector))
```

tbruyelle:
@bradleypeabody Thanks this is fixed. Can you tell me why we need to add those 4 extra bytes for strings ?

bradleypeabody:
@tbruyelle Thanks and this is now merged!

Strings are encoded as a 32-bit big endian length prefix followed by the number of bytes indicated.  https://github.com/vugu/vugu/blob/master/domrender/renderer-js-instructions.go#L448 . So "AB" is encoded as 6 bytes: 0x000000024142

The encoding scheme was designed to be simple to write in Go and read in JS, and have a length that can be known up front and not require escaping. There are only a few value types so bringing in a protobuf or even msgpack encoder/decoder seemed overkill.  These values take only a couple lines of vanilla JS to implement the decoder.

PR #86, opened 2019-12-08
Fix static fields always considered as attrMap

### LINKED ISSUE #85: Static fields of components are always considered as attrMap
**Describe the bug**
A component field with an upper case should be linked to a component regular field, but it's currently considered as an `attrMap` component (like if it started with a lower case).
This results to a compilation error if no `attrMap` field has been declared in to component.

**Software Versions**
Vugu version: fa72e903246bf90a0b548f7bb4dd607b9d60f4d4
Go version: 1.13.4
Browser and version: Version 78.0.3904.108 (Official Build) Arch Linux (64-bit)

**To Reproduce**
Display a component and feed it with a static regular field of the component.

(A MR will follow with a complete test case)

**Expected behavior**
A static field with upper-case first letter should match a component regular field.

tbruyelle:
Fix #85

The fix is straightforward. Because the field name case is important since
the last refactor, we should always deal with attr.OrigKey and never attr.Key.

Note: we could remove `attrFromHtml` func and inline it inside `staticVGAttr`, since it's only used here.

### DIFF
--- a/gen/parser-util.go
+++ b/gen/parser-util.go
@@ -19,7 +19,7 @@ import (
 func attrFromHtml(attr html.Attribute) vugu.VGAttribute {
 	return vugu.VGAttribute{
 		Namespace: attr.Namespace,
-		Key:       attr.Key,
+		Key:       attr.OrigKey,
 		Val:       attr.Val,
 	}
 }
--- a/wasm-test-suite/test-007-issue-85/.gitignore
+++ b/wasm-test-suite/test-007-issue-85/.gitignore
@@ -1,0 +1,7 @@
+main_wasm.go
+main.wasm
+root.go
+comp.go
+go.sum
+wasm_exec.js
+index.html
--- a/wasm-test-suite/test-007-issue-85/comp.vugu
+++ b/wasm-test-suite/test-007-issue-85/comp.vugu
@@ -1,0 +1,7 @@
+<p vg-html='c.Key'></p>
+
+<script type='application/x-go'>
+  type Comp struct {
+    Key string `vugu:"html"`
+  }
+</script>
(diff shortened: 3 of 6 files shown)


bradleypeabody:
Okay great and thanks.  Please add comp.go to the .gitignore in that same folder and remove it from the PR.  Otherwise this looks good.

tbruyelle:
@bradleypeabody ah yes I forgot it. It's removed and ignored now.

Do you consider as a good vugu practice to ignore component generated code ? I did not for the moment in my project, I find a little boring and error prone to repeatedly add new component to `.gitignore`.

One solution could be to generate code into more *recognisable* files like `*-gen.go`, and so we can simply add that pattern in `gitignore`. That would also leave the user the possibility to move its go code in `comp.go` file, because since the refactor, using `comp-data.go` file for go code became weird.

bradleypeabody:
@tbruyelle Thanks, merged.

> Do you consider as a good vugu practice to ignore component generated code
> ...
> One solution could be to generate code into more recognisable files like `*-gen.go`

The decision to ignore generated .go files depends on if the package is intended by used by others or not.  When you are writing your own application ignoring generated .go files makes sense.  But if you are making a reusable component package that you don't want to have to keep re-generating, then it makes sense to have the generated go files committed.  This generally follows other uses of go generate - it is run by the package maintainer, but not by the users of that package.

I agree a better naming convention would make sense.  `-gen.go` as a suffix seems pretty logical.  I've seen some people use an underscore instead but I dislike the possible confusion of it looking like or actually being used as a build tag.

Another related change would be to improve the logic on the decision to clobber a generated file.   A simple change would be to check its contents for the "DO NOT EDIT" warning.  And even better would be some sort of checksum which maybe could go in a comment and be used to tell if someone has changed the file.  The logic on that last could be used for main_wasm.go as well - which we don't overwrite because we don't want to clobber people's changes - but if we can tell if any hand edits have been made, it should be fine to update.

Issue #89, opened 2019-12-10
vg-for and events: value and key are not captured


tbruyelle:
**Describe the bug**
Inside a loop, when using `value` or `key` variable inside a event attribute (like `@click`), the last value of `key/value` of the loop is always used, because they are not captured.

**Software Versions**
Vugu version: d1cd4bcee583c7d794f2d1278962e98662b39a6e
Go version: 1.13.5
Browser and version: Brave Version 1.0.0 Chromium: 78.0.3904.97 (Official Build) unknown (64-bit)

**To Reproduce**
A PR will follow with a failing test case.

**Expected behavior**
`key/value` inside event attribute should keep their value at the moment they are used.

**Additional Notes**
There's a TODO for that, but I have no idea on how to achieve that properly :
https://github.com/vugu/vugu/blob/master/gen/parser-go.go#L794

bradleypeabody:
Interesting - good point.

This one needs some careful thought.

Generally it seems fine to have two sets of variables, one from the for statement and one that is inside the loop's scope.  But there is the question of how to name these so they are sane and consistent and also how to make it clear to the user which is which.

Logically if I put variable names in vg-for I am going to, as a user of the library, want to reference those variable names later.  Having some other alternate name that I can't see which means the version that is captured inside the loop - I feel like that's going to be confusing.

Ideally whatever syntax and names are used would:

- Translate in an obvious way to a Go `for`
- Allow the user to choose variable names used in the `for` statement but provide sensible defaults as a shorthand
- We need vgiterkey to be defined sensible in order for caching to work

Possibly another attribute could be introduced which gives variable declarations inside the loop, e.g. `<div vg-for="k, v := range w" vg-var="k2 := k; v2: = v">` (the vg-var stuff is emitted inside the loop).

I will think about this more. I'm definitely open to suggestions.

tbruyelle:
Indeed I manage to provide a fix for when the user doesn't provide the variable name (so by default he knows this implies to use `key` and `value`), but this doesn't work for the other case, when he provides the variable name.

Maybe we could *cheat* on the variable names, for instance if the user declares

```html
<div vg-for='k, v := range w'>
```

the generated code is

```go
for k_, v_ := range w {
  k, v := k_, v_ 
 ....
}
```

What do you think ?

bradleypeabody:
I'm still thinking about this, but have a look at this example: https://play.golang.org/p/FSbh6-qOZtI

[...]

PR #90, opened 2019-12-10
Fix non captured loop variables

tbruyelle:
Related to #89

See https://github.com/vugu/vugu/blob/master/gen/parser-go.go#L794

This is a quick fix, that probably requires embellishment. I'm not 100% sure it the best way to do it.


tbruyelle:
I finally found a way to capture `key/value`, but that said I'm not 100% sure it the best way to do it.

bradleypeabody:
@tbruyelle  It's great that you're working on this.  Let me know when it's ready for me to review again.

tbruyelle:
Yes, I still need to implement the `noshadow` option and I let you know.

tbruyelle:
@bradleypeabody I want your opinion about 2 possible implementations for handling the options in attributes. I think it's better to have a implementation that can be used for all vugu *directives* (to borrow the vue.js term), a function like `vgDirective` or `vgAttr`, that takes the node and the directive name. 

### simple implementation

A simple implementation returns as before the value of the attribute, and a second value that represents the options.
```go
func vgAttr(*html.Node, attr string) (value string, options map[string]bool)
```
This has the benefit of impacting a minimum amount of code, and consumer can easily ignore the options by assigning them to `_`.

### robust implementation

A more robust implementation would rely on a new specific type. 
```go
func vgAttr(*html.Node, attr string) VGAttr // I don't like this name
```
This type holds the attribute key and value, and has methods like `Value() string` to return the attribute value or `HasOption(string) bool` to say if an option is set. The benefit is it's easier to read the options and the code is more readable. It's also more flexible (we could add later `GetArg() string` to represent directive args, like vue.js again) but it requires more changes in the code base (since it's no longer possible to access the value directly).

Which one do you prefer ? (on my side it's the robust one :))

bradleypeabody:
@tbruyelle I generally agree with what you're saying about making a robust version.  However, I'm not sure there is enough of a pattern here to determine the best way to do that.  Note that the `vgAttr` function is used internally and only during code generation. Where as things like `VGNode` are used at runtime.  So I'm not sure that a `VGAttr` type is appropriate.  I think what you're describing is a code-generation-time concept, not runtime.

If the type you want to make is not exported and only used in the `gen` package, then it's probably fine.  But again, I'm concerned about trying to form a pattern when we only have one example so far.  My spidey-sense says to keep it simple enough we understand the use cases better.

tbruyelle:
@bradleypeabody Indeed, thanks for the clarification. I had time yesterday so I experimented the robust implementation. During this I also decided to use only private types. You can have a look 37333d89ed561ceb780e5eeb0e8eeca3788dad8f, note that this is unfinished work, the naming is rough and not all I only used `vgAttr` in `vgForExpr`, and not yet in the other `vg*Expr`. 

Tell me if it's simple enough for you, otherwise I can re-implement it in an other way.

tbruyelle:
@bradleypeabody I read again you answer, and maybe the simplest way would be to only add a function like 
```go
func vgAttrOption(n *html.Node, attr, option string) bool
```
that returns true if the specified option is present for the attribute. Thus we don't change the signature of other functions responsible of retrieving the attribute value, this has less impact on the code.

That said a small change is still required for `vgForExpr`, instead of comparing  `a.Key == "vg-for"`, we must have `strings.HasPrefix(a.Key, "vg-for")` because of the potential option. Question is should we apply that kind of comparison for all vugu directives, even if they don't have options for the moment ?

bradleypeabody:
@tbruyelle What about this approach:

```
func vgForExpr(n *html.Node) vgForAttr { /* ... */ }

type vgForAttr struct {
    Expr string // the expression, empty string means not present
    Noshadow bool // true if the "noshadow" option was specified
}
```

This way should be easy to adapt existing code and `vgForExpr` can just parse exactly what it needs.  I do not think it should allow anything except for the exact cases that are valid `vg-for=` and `vg-for.noshadow=` are okay, `vg-for.blah=` should fail.   The parsing on this would need to be more fancy if multiple options were supported, but since that logic would be localized in `vgForExpr` it really doesn't matter as it's trivial to change later in one place.

Let me know if that makes sense.

tbruyelle:
@bradleypeabody got your point, it makes totally sense. I'll go for that as soon as I have time.

tbruyelle:
@bradleypeabody Small question about the error part (when an option isn't recognized), the signature of the function you described returns no error, does that mean you want the code to panic ? Or should I add an error in the returned values ?

bradleypeabody:
@tbruyelle Please add an error return from that function.

tbruyelle:
@bradleypeabody the MR is ready for review

gen/parser-util.go
```diff
@@ -98,21 +99,30 @@ func vgKeyExpr(n *html.Node) string {
...
-			// }
+type vgForAttr struct {
+	expr     string
+	noshadow bool
+}
 
-			return v
+func vgForExpr(n *html.Node) (vgForAttr, error) {
+	for _, a := range n.Attr {
+		if strings.HasPrefix(a.Key, "vg-for") {
```
bradleypeabody: This will also match `vg-forabcdef`, better would be to change it to `a.Key == "vg-for" || strings.HasPrefix(a.Key, "vg-for.")`.  But I will go ahead and merge and just do this one change after.

bradleypeabody:
Great work!  (Sorry for the delay in reviewing - holidays have been crazy).

Merging.

tbruyelle:
@bradleypeabody thx, no problem for the delay I was in holidays too :) Happy new year!

Issue #101, opened 2020-01-18
Rendering a collection with vg-for shows the wrong number of elements when the size of the collection decreased after an update.


mazzegi:
**Describe the bug**
Rendering a variable size collection with `vg-for` shows the wrong number of elements when the size of the collection decreased after an update. 

**Software Versions**
Vugu version: v0.1.1-0.20191230174203-fe17e20eb6ce
Go version: go1.13.4 linux/amd64
Browser and version: Chrome, Version 79.0.3945.117 (Official Build) (64-bit)

**To Reproduce**
The .vugu file below illustrates the issue:

```
<div>
    <button @click='c.Make(2, event)'>Make 2</button>
    <button @click='c.Make(6, event)'>Make 6</button>

    <div vg-for='_,n := range c.Numbers' vg-key='n' >            
        <div vg-html='n'></div>        
    </div>
</div>

<script type="application/x-go">
import(
    "fmt"
    "log"    

    "github.com/vugu/vugu"
)

type Root struct {       
    Numbers []string `vugu:"data"`
}

func (c *Root) Make(cnt int, event *vugu.DOMEvent) {
        event.PreventDefault()	
	go func() {        
		ee := event.EventEnv()
		ee.Lock()
		defer ee.UnlockRender()

               log.Printf("show %d", cnt)
		c.Numbers = make([]string, cnt)
		for i := 0; i < cnt; i++ {
			c.Numbers[i] = fmt.Sprintf("Number %d", i+1)
		}
	}()
}
</script>
```

When clicking "Make 2", two items are shown.
When clicking "Make 6", six items are shown.
When clicking "Make 2" again, five items are shown.

**Expected behavior**
When clicking "Make 2", two items are shown.
When clicking "Make 6", six items are shown.
When clicking "Make 2" again, two items are shown.

bradleypeabody:
Thanks for reporting this.  I'm looking into it and will post any updates as soon as I have them.  Might be related to https://github.com/vugu/vugu/issues/99

itstmyi:
Any new update about this issue ? Thanks .

jyellick:
FWIW, I'm running into this as well.

bradleypeabody:
@mazzegi @jyellick I just pushed a fix for this to master.  Can you please try again with the latest code?

mazzegi:
Just gave it a quick try. At least the illustrating example works now.

bradleypeabody:
Excellent.  I think we're good on this.

PR #105, opened 2020-03-03
Track unused elements and cleanup at end

jyellick:
There is a bug, as nicely documented with example in:

https://github.com/vugu/vugu/issues/101

Where vg-for ranges do not appropriately delete elements when the range
shrinks.

This happens because the DOM rendering re-uses the parent node, and
re-uses a subset of the existing child elements, but does not delete
the remaining child elements.

This commit adds a field which tracks the children of any re-used
elements, deleting them as they themselves are re-used, and finally
removes those which were never re-used at the end of the render.

Signed-off-by: Jason Yellick <jyellick@us.ibm.com>


jyellick:
This seems to erroneously flag some elements as removed, causing unnecessary churn in the DOM -- I would not merge as is, but thought it generally illustrates what I think is the underlying problem.

jyellick:
Added an additional check that all re-used elements are in fact found as children in the set.  So I'm at a loss to explain some of the elements I show as being deleted.  Things still seem to render correctly, perhaps @bradleypeabody will have some more insight.

jyellick:
Found the problem -- I was taking opcodeEnd to mean the render had ended, but it is in fact only the current buffer.  There was no obvious opcode that said "render done" to me, but `opcodeRemoveOtherCSSTags` seems to be the last thing called on every render, so stuck the cleanup there for the moment.  Feels like there probably ought to be a 'render done' opcode to generally handle cleanup like this, but didn't want to do anything too invasive, so left the hacky insertion into the css tags for the moment.

bradleypeabody:
@jyellick Thanks for the work you've done on this - definitely looks like it took a fair bit of effort to track down and come up with this fix.  I wanted to acknowledge that right out of the gate.

You're right that opcodeEnd is the end of the buffer, not necessarily the whole render itself.   It's possible something could be added for this but I'm not sure it's the best way to go. When I wrote this code I was of the mindset that I wanted the instructions to have exactly defined relationships to each other, rather than understanding what a "render cycle" is.   For example, the instructions related to CSS tags have behaviors that are consistent between each other, but they do not understand anything outside of that.  I felt like adding things for "end of render cycle" would tend to accumulate edge cases rather than forcing more thought to be put into the relationship between different instructions.

I'm a little concerned about the performance implications of the approach you're using.   It requires walking the entire DOM tree and collecting up all of the nodes as a first step (I'm also not sure if this code handles text and comment nodes correctly).  If there is a way to accomplish what you're doing but without walking large chunks of the DOM tree up front I think it would be better.

I'm still looking at this but that's my initial feedback.

bradleypeabody:
After going through this in more detail it seems like the fix is actually fairly simple, although admittedly not obvious.

The DOM elements are meant to be traversed in sequence, depth first.  Any new elements are created, existing ones are overwritten, and the missing piece was to remove any which do not exist in the new virtual DOM we're using for our updates.

So I did a fix which removes any other sibling nodes whenever we move up to the parent node.  This works because, by definition, if we traverse in sequence depth first, then when we move up to a parent it is implied that there are no more siblings for the current level - thus we can remove any that are there in the DOM.  The relevant commit is here: https://github.com/vugu/vugu/commit/0f4714151aaa7bc1c2025b381c568d7c6a9bd03a

I think that resolves the issue from #101 and makes this PR unnecessary.

Let me know if you feel we need to discuss this further.  Again, I do appreciate the effort you put in on this PR.  You also correctly identified the specific problem which was a huge help in resolving it.

jyellick:
No problem at all, I suspected the above solution was overkill.  Thanks for providing the correct fix!

PR #112, opened 2020-03-24
Add buffered set inner html opcode

jyellick:
When a user tries to have the framework set a large HTML element (for
instance, large blocks of preformatted text), if the text itself exceeds
the buffer size minus some overhead (today, 16384 bytes), then the
application panics with 'requested instruction does not fit in the
buffer'.

This commit adds a new opcode opcodeBufferInnerHTML which can be called
in succession to buffer a large block of innerHTMl over multiple calls
on the JS side.  On the golang side, we loop over the innerhtml, so long
as it would overflow our buffer, we chunk up the inner html over
multiple calls to this opcode.  Finally, when the original setInnerHTML
op is performed, we prepend the buffer, if is not empty.


bradleypeabody:
Thanks Jason.  My only thought on this is that so far the `write` methods to correspond to exactly one instruction.  The change breaks that rule.  That said, it's the only case so far and it has no functional impact, so I'm okay with it for now.  If more cases of multiple instructions come up like this it might make sense to call these ones something other than `write`..., but it's not worth fussing about until there are more cases to take into account.

I will look at this in just a bit more detail within the next day or so and let you know if there are any other concerns.  But otherwise I generally like it and will merge.

Thanks for the great work!

Issue #121, opened 2020-04-10
There should be a way to properly handle `disabled` and other HTML attributes that have no value and/or need to be conditionally output


bradleypeabody:
**Is your feature request related to a problem? Please describe.**

There is currently no way to express the following markup with a condition for the `disabled` attribute to be shown or hidden based on a Go expression:

```html
<div><button disabled>test</button></div>
```

This is because the only option for dynamic attributes is the `:` syntax but this always outputs the attribute.

**Describe the solution you'd like**

This is still open for discussion but the current idea being considered is to essentially allow the colon syntax with e.g. `:disabled="...` to hide attributes where the Go value evaluates to `false` or `nil` based on type, more specifically:

- `string` - value is used directly as attr value
- `[any number type]` - converted to a string using applicable `strconv` func
- `bool` - if false does not output attribute, if true outputs it with no value.
- `[pointers to these types]` - will be followed and the value used, if nil then the attribute hidden as if it were `false`
- (see explanation below about `VGAttributeList() []VGAttribute`)
- `String()` - if it implements `String() string` then we call that method and use it as if it were a `string`
- anything else is undefined behavior but currently implemented with fmt.Sprint

We should also consider if there are one or more interfaces here that could provide a good way for arbitrary types to express how they want to be translated to an attribute.  E.g. `type VGAttributeLister interface { VGAttributeList() []VGAttribute }` - we could check if a type implements that interface and use it for attribute conversion in that case.  This should take precedence over the String() method if the type has both.  Interestingly, this also means there could potentially be a `VGAttributeListerFunc` type which (following the pattern of `http.HandlerFunc`) makes it easy to write custom code that emits whatever attributes it wants.  It's a bit unclear what the attribute name means on these VGAttribute structs returned - since there is already a name specified in the `:disabled="...` part.  Possibly returning `VGAttribute{Val:"whatever"}` with an empty attr name would get the name `disabled` from the attr in the source, whereas if you specify an attr name in the return it is used.  This is TBD, but overall this could be really useful and be both convenient for simple cases but also handle complex ones well.

[...]

PR #122, opened 2020-04-11
Better dynamics attributes

tbe:
Implementation of #121

### DIFF
--- a/attribute-lister.go
+++ b/attribute-lister.go
@@ -1,0 +1,15 @@
+package vugu
+
+// AttributeLister implements the required functionality, required by the `:=` attribute setter.
+// Types may implement this interface, to
+type VGAttributeLister interface {
+	AttributeList() []VGAttribute
+}
+
+// The VGAttributeListerFunc type is an adapter to allow the use of functions as VGAttributeLister
+type VGAttributeListerFunc func() []VGAttribute
+
+// AttributeList calls the underlying function
+func (f VGAttributeListerFunc) AttributeList() []VGAttribute {
+	return f()
+}
--- a/vgnode.go
+++ b/vgnode.go
@@ -1,5 +1,11 @@
 package vugu
 
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+)
+
 // Things to sort out:
 // * vg-if, vg-range, etc (JUST COPIED FROM ATTRS, EVALUATED LATER)
 // * :whatever syntax (ALSO JUST COPIED FROM ATTRS, EVALUATED LATER)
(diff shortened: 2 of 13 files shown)


attribute-lister.go
```diff
@@ -0,0 +1,7 @@
+package vugu
+
+// AttributeLister implements the required functionality, required by the `:=` attribute setter.
+// Types may implement this interface, to
+type VGAttributeLister interface {
+	ListAttributes() []VGAttribute
```
bradleypeabody, line 6: Let's call the method `AttributeList` to go along with the convention of "interface `Aer` has method `A`".  (The "VG" prefix is a little funky - all of this VGNode and VGAttribute might move to a separate package at some point in order to get rid of that prefix, but I don't want to get into that now.)

vgnode.go
```diff
@@ -162,25 +168,107 @@ func (n *VGNode) RemoveChild(c *VGNode) {
...
+		}
+
+		// fall back to fmt
+		nattr.Val = fmt.Sprintf("%v", val)
+	}
+
+	n.Attr = append(n.Attr, nattr)
+}
+
+// SetAttrList takes a VGAttributeLister and sets the returned attributes to the node
```
bradleypeabody, line 265: Looking at this newly, I think `Set` is not the right word here (for both SetAttrList and SetAttrInterface) since they both append().  I would expect a `Set` method to overwrite any existing value.  Please change `Set` to `Add`, I think that will be more clear - AddAttrList and AddAttrInterface.

vgnode.go
```diff
@@ -162,25 +168,107 @@ func (n *VGNode) RemoveChild(c *VGNode) {
...
 }
 
+// SetAttrInterface sets an attribute based on the given interface. The followings types are supported
+// - string - value is used as attr value as it is
+// - int,float,... - the value is converted to string with strconv and used as attr value
+// - bool - treat the attribute as a flag. If false, the attribute will be ignored, if true outputs the attribute without a value
+// - fmt.Stringer - if the value implements fmt.Stringer, the returned string of StringVar() is used
+// - ptr - If the ptr is nil, the attribute will be ignored. Else, the rules above apply
+// any other type is handled via fmt.Sprintf()
+func (n *VGNode) SetAttrInterface(key string, val interface{}) {
```
bradleypeabody, line 197: change to `AddAttrInterface`, see below.

attribute-lister.go
```diff
@@ -0,0 +1,7 @@
+package vugu
+
```
bradleypeabody, line 2: How about adding:

```go
type VGAttributeListerFunc func() []VGAttribute
func (f VGAttributeListerFunc) ListAttributes() []VGAttribute { return f() }
```

This will allow things like `<div :='vugu.VGAttributeListerFunc(func() []VGAttribute { return ... } )' >`

(Same pattern as [http.HandlerFunc](https://golang.org/pkg/net/http/#HandlerFunc))

bradleypeabody:
I'm torn as to whether or not `:=` should be `vg-attr=` - it would be more consistent with rest of the naming.  I will probably merge as-is and then add vg-attr= support and in the doc say that `:=` is a shorthand for `vg-attr=`.  Still thinking about this, just making the note so I don't forget.

tbe:
After a second and third thought:

IMHO: As attributes could be anything, any possible naming conflict should be avoided. And even it may be never required, we should allow the user to use `vg-attr` as an attribute name.

bradleypeabody:
We already have to assume that the user will never need to use `vg-` for their attributes because of `vg-for`, `vg-if`, etc.  `vg-` is reserved for Vugu. By that logic I don't think it's any more imposing than what Vugu already does.

tbe:
rebased and squashed the changes. Reworded the commit message.

bradleypeabody:
Thanks again for the excellent work on this!

Issue #123, opened 2020-04-11
Allow the creation of namespaced elements


tbe:
**Is your feature request related to a problem? Please describe.**
Some commonly used objects, like `svg` must be created within a namespace. Currently this is not supported in `vugu`. It requires the usage of `createElementNS` in JS.

**Describe the solution you'd like**
After a rough look over the code, the following is missing
- create a new `opcodeSetElementNS`
- implement the js code to call `createElementNS`
- check if a node contains a namespace and set the correct opcode

`VGNode` already has a namespace attribute, but as far as i have seen it is not used right now. 

**Describe alternatives you've considered**
One could always use `syscall/js` but this, if at all, a bad workaround.

bradleypeabody:
There is a note in the parser code to add a check for upper case to distinguish `<pkg:Comp>` from `<ns:tag>`.  Checking for the capital letter would be an easy way to distinguish namespaced tags from component references.

Then, I wonder if we really need a new opcode or if the attribute is called "svg:line" then we could check for the colon in the opcodeSetElement code and call createElementNS there as needed.  I'm actually not clear myself on what conditions cause the [Go HTML parser](https://godoc.org/golang.org/x/net/html) (which we use a fork of for Vugu) to populate the `Namespace` field instead of `Key` on a Node.  I think with a clear understanding of that we could see what the best way is of piping that information into the UI is.

tbe:
Just seen, it is not that easy at all. `createElementNS` requires the namespace as URL. So, for `svg` that would be `http://www.w3.org/2000/svg`.

For `svg` and `math` ( supported by the HTML parser out of the box ), the namespace URL's could be provided by vugu from an internal map.

Leaves two questions:
- How does a user provide other namespace URLs?
- Should the mapping short NS -> URL be done in JS or Go?

bradleypeabody:
Makes sense.  If you can fiddle with the Go HTML parser and see how it handles different situations, this would help.  Like I don't think I've ever seen the `Namespace` field not-empty in any HTML I've ever parsed with that thing.  I'm not sure if it's supposed to be the full namespace like `http://www.w3.org/2000/svg` or just `svg` or what.  I think that would be the first thing to clarify.

tbe:
From Slack discussion:
- the HTML parser sets `svg` or `math` as `Namespace` on elements that are part of these namespaces. Else the `Namespace` attribute is always empty

bradleypeabody:
This is all done and merged now.

PR #124, opened 2020-04-12
Basic support for namespaced elements

tbe:
Some HTML elements are located in a different namespace.
This commit implements support for the namespaced defined in the
HTML specification.

The supported namespaces per HTML spec are:
- html
- math
- svg
- xlink
- xml
- xmlns

see #123

Signed-off-by: Thomas Berger <loki@lokis-chaos.de>


domrender/renderer-js-script.js
```diff
@@ -1,7 +1,8 @@
+(function () {
```
bradleypeabody, line 1: FYI I'm not a big fan of formatting the JS file because there's no standard and if multiple people do it it can be more hassle than it's worth.  But this once is fine.

vgnode.go
```diff
@@ -32,7 +32,7 @@ const (
 
 // VGAttribute is the attribute on an HTML tag.
 type VGAttribute struct {
-	Namespace, Key, Val string
+	Namespace, NamespaceURI, Key, Val string
```
bradleypeabody, line 35: This concerns me - VGAttribute is copied from https://godoc.org/golang.org/x/net/html#Attribute and I want to consider carefully before changing it.

If the NamespaceURI can be derived from `namespaceToURI`, why do we need to store it?

vgnode.go
```diff
@@ -47,9 +47,10 @@ type VGNode struct {
 
 	Type VGNodeType
 	// DataAtom  VGAtom // this needed to come out, we're not using it and without well-defined behavior it just becomes confusing and problematic
-	Data      string
-	Namespace string
-	Attr      []VGAttribute
+	Data         string
+	Namespace    string
+	NamespaceURI string // required by the domrender for createElementNS
```
bradleypeabody, line 52: same issue here

domrender/renderer-js-script.js
tbe: In fact, there is ;) https://standardjs.com/

And sorry, me IDE didn't use it :(

PR #127, opened 2020-04-17
Handle dynamic attributes correctly in all places

tbe:
this implements #122 for all locations where dynamic attributes are
supported. In the initial commit, i missed some places where these kind
of attributes are handled.


bradleypeabody:
Looks good to me, thanks for fixing.

PR #156, opened 2020-08-23
event listeners are not removed

san-321:
test case
```
<div>
  <input type="button" @click='c.b=!c.b' value="toggle">
  <input type="button" vg-if="c.b" value="no click"><input type="button" @click='println("click")' value="click">
</div>

<script type="application/x-go">
type Root struct {
    b bool
}
</script>

```
After clicking the "toggle" button on the "no click" button there will be a click listener.
Аfter the next "toggle" click there will be two listeners on the "click" button.
etc


bradleypeabody:
Thanks for this, interesting issue.

If you're able to add a test case for this (see https://github.com/vugu/vugu/tree/master/wasm-test-suite) which demonstrates the issue and shows that this fix makes it work properly, that would help confirm this is the right fix.

san-321:
added test

bradleypeabody:
Thanks for making the test for this!  Overall it seems fine but I think I'm still just dumbfounded about why `{capture: +kparts[1], passive: +kparts[2]}` works when using `!!` doesn't...  Per https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener `capture` and `passive` are supposed to be booleans.  Do you have any other info on why this works?

san-321:
kparts[1], kparts[2] are strings ("0" or "1").
And in javascript "0" is converted to boolean true, although 0 to false.
Therefore !!"0" === true, and +"0" === 0 == false.

However "0" == false, because both sides of the expression are converted to a number :confused:

bradleypeabody:
@san-321 Thanks, merged!

PR #194, opened 2021-04-08
reallocate instructional buffer size when it's needed

mertnacakgedigi:


bradleypeabody:
@mertnacakgedigi Thanks for this and it generally looks like it's on the right track. I think it needs a test case to verify.

domrender/renderer-js-instructions.go
```diff
@@ -123,12 +123,30 @@ func (il *instructionList) checkLenAndFlush(l int) error {
 				return err
 			}
 			err = il.checkLen(l)
+			if err == errDoesNotFit {
+				size := computeRequiredSize(il.pos + l)
```
sorenisanerd, line 127:
```suggestion
				size := computeRequiredSize(l)
```
After a succesful call to il.flush(), il.pos will be 0.

domrender/renderer-js-instructions.go
```diff
@@ -123,12 +123,30 @@ func (il *instructionList) checkLenAndFlush(l int) error {
...
+				buf := make([]byte, size)
+				copy(buf, il.buf)
+				il.buf = buf
+				err = il.checkLen(l)
```
sorenisanerd, line 131: I'm not sure how this would fail.

sorenisanerd:
```suggestion
```

sorenisanerd:
```suggestion
				return nil
```

Issue #192, opened 2021-04-03
vugufmt applies ToLower on HTML if used recursively 


dbd:
**Describe the bug**
Running `vugufmt -d components/my-line.vugu` standard html diffs. Running `vugufmt -d .` will try to run ToLower on the same HTML.

**Software Versions**
Vugu version:  v0.3.4
Go version: 1.15.8
Browser and version: N/A

**To Reproduce**
```
# vugufmt -d components/my-line.vugu
# vugufmt -w components/my-line.vugu
# vugufmt -d .
diff -u root.vugu.orig root.vugu
--- root.vugu.orig      2021-04-03 10:25:17.110000000 -0500
+++ root.vugu   2021-04-03 10:25:17.110000000 -0500
@@ -1,7 +1,7 @@
 <div class="root">
-    <ul>
-      <components:MyLine FileName="example.txt" :LineNumber="rand.Int63n(100)" ></components:MyLine>
-    </ul>
+       <ul>
+               <components:myline filename="example.txt" :linenumber="rand.Int63n(100)" ></components:myline>
+       </ul>
 </div>

 <script type="application/x-go">
```

**Expected behavior**
HTML is only formatted and converted to lower case.

**Additional Notes**
example repo: https://github.com/dbd/vugu-fmt-bug

bradleypeabody:
Thanks for the bug report.  Yeah vugufmt needs some love, it hasn't been updated in a while.  I'll leave this open for now.

Issue #154, opened 2020-08-17
vugufmt breaks on case sensitive input (vgform:Input, vgform:Select)


seanrmurphy:
**Describe the bug**
vugufmt does not handle `vgform:Input` and `vgform:Select` properly - the resulting output loses the upper case letter after the colon.

**Software Versions**
Vugu version: v0.3.2-5-g1845706
Go version: go version go1.14.6 linux/amd64
Browser and version: Not relevant.

**To Reproduce**
Create a `.vugu` file which contains a `vgform:Input` element and run it through `vugufmt`

```
<div>
        <form>
                <div>
                        <label>Input Data:</label>
                        <vgform:Input type="string" :value='vgform.StringPtr{&c.InputString}'
                                ></vgform:Input>
                </div>

        </form>
</div>


<script type="application/x-go">

import "github.com/vugu/vugu/vgform"

type InputDataExample struct {

	InputString string
}

</script>
```

**Expected behavior**
Output:

```
/tmp> vugufmt input-data-example.vugu
<div>
	<form>
		<div>
			<label>Input Data:</label>
			<vgform:input type="string" :value='vgform.StringPtr{&c.InputString}'
                                ></vgform:input>
		</div>

	</form>
</div>


<script type="application/x-go">

import "github.com/vugu/vugu/vgform"

type InputDataExample struct {
	InputString string
}

</script>
```

`vgform:Input` should not have been modified to `vgform:input`

**Additional Notes**
None.

seanrmurphy:
Also impacts `:Value` and `:Options` - should have flagged this in the above.

bradleypeabody:
Thanks for this.  Yeah the vugufmt code is a bit out of date, needs some TLC.  I'll leave this open to address the next time maintenance is done on that tool.

owenwaller:
Closing this as the problem seems to be resolved by #195.

PR #195, opened 2021-04-09
remove lower() call on tokens with colons

dbd:
This would resolve the vugufmt lowering mentioned in #192 and #154 (apologies for not looking more before creating my own issue).

The two calls to TagName still use the same code path that they were before in my testing. The related tests also still passed.


bradleypeabody:
Okay great!  Looks good, merging.

PR #205, opened 2021-07-17
Fix "ValueOf: invalid value" panics in syscall/js when using vugu/js

Dadido3:
This is a straight forward fix for the problem that syscall/js doesn't know about vugu/js value.

### DIFF
--- a/js/impl-js.go
+++ b/js/impl-js.go
@@ -109,7 +109,7 @@ func Global() Value {
 
 // ValueOf alias to syscall/js
 func ValueOf(x interface{}) Value {
-	return Value(sjs.ValueOf(x))
+	return Value(sjs.ValueOf(fixArgToSjs(x)))
 }
 
 // CopyBytesToGo alias to syscall/js
@@ -172,7 +172,7 @@ func (v Value) Get(p string) Value {
 }
 
 func (v Value) Set(p string, x interface{}) {
-	sjs.Value(v).Set(p, x)
+	sjs.Value(v).Set(p, fixArgToSjs(x))
 }
 
 func (v Value) Index(i int) Value {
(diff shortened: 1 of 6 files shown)


bradleypeabody:
Thanks for this.  Is there a test case that you can add to the test suite in `wasm-test-suite`, so I can see what problem is being solved with this fix?

Dadido3:
I haven't got the wasm-test-suite to run without errors here. Any tests always fail with `panic: context deadline exceeded` on the line `must(chromedp.Run(ctx, actions...))`. The docker container is running, maybe there is something wrong with my chromium installation.

Anyways, the problem that this pull request fixes can be simple reproduced by this code:
```go
import js "github.com/vugu/vugu/js"

func main() {
	img := js.Global().Get("Image").New()
	img.Set("src", js.ValueOf("someURL")) // js.ValueOf() isn't really needed here, but it shows the problem.
}
```

If i get the wasm-test-suite working the next few days, i'll write a test for it.

Dadido3:
Finally got the tests to run in a linux container, and i added a test for the fix. Hope this covers the problem well enough. What's not covered by this test is:

- `FuncOf(...)`
- `ValueOf(...)`
- `SetIndex(...)`

But the fix works similar for these functions.

I've also figured out why the tests didn't run under windows, i'll make another pull request for that fix.

sorenisanerd:
The tinygo build should work when #219 is merged.

Dadido3:
> a) can you confirm or not that this is the same issue as #328 (I'm assuming not ATM)
It's the same issue, but my fix wasn't complete, so there were some other `vugu/js` functions that still caused the same panic.

Closing due to newer pull request #331 with a more complete fix.

PR #206, opened 2021-08-13
Fix relative paths in mustUploadDir

Dadido3:
Generated tar archive names were not independent of OS paths. Now they are, and it's possible to run the tests on windows machines.

### DIFF
--- a/wasm-test-suite/wasm-suite-util_test.go
+++ b/wasm-test-suite/wasm-suite-util_test.go
@@ -15,7 +15,6 @@ import (
 	"mime/multipart"
 	"net/http"
 	"os"
-	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


wasm-test-suite/wasm-suite-util_test.go
```diff
@@ -265,14 +264,20 @@ func mustUploadDir(dir, endpoint string) string {
 
 		absPath, err := filepath.Abs(fpath)
 		if err != nil {
-			panic(err)
+			return err
```
sorenisanerd, line 267: AFAICT, the `must*` functions are expected to succeed or panic.

Dadido3: It does already. The error is returned by `filepath.Walk()` and will trigger a panic some lines below.

But `panic(err)` may be easier to read, though. It probably would be best to change all occurences of `return err` into `panic(err)` inside of the `WalkFunc`. (Or leave them all at `return err`. Mixing both variants is a bit strange)

Dadido3:
This fix is not applicable anymore since the old wasm-test-suite will be removed in the future.

PR #207, opened 2021-08-13
Add missing syscall/js functions

Dadido3:
- Add `value.Delete(...)`
- Add `value.Equal(...)`
- Add `value.IsNAN()`
- Prevent values from being compared, similar to how `syscall/js` does it.


Dadido3:
Closing due to newer pull request #331 with a more complete fix.

PR #218, opened 2022-01-27
Go 1.16 and 1.17 support

sorenisanerd:


bradleypeabody:
Hi @sorenisanerd Can you please summarize (explain) the key changes?  It will be easier to merge if we break some of these things down.  I like the idea of moving to GitHub actions but I think that could just be one PR and we do that as one thing.  On some of the other changes I don't follow for example why "go mod tidy" is being run in these various places, or what the purpose of some of other changes is.

sorenisanerd:
Yeah, sorry, didn't mean to include the GitHub Actions thing in this PR. I'll get rid of it and write a better commit message :)

sorenisanerd:
It doesn't make a whole lot of sense to not have the travis -> actions change here, since I have to change the test matrix, but I've sent a PR for it, so it can be reviewed first and separately.

Also updated the description of this PR.

PR #219, opened 2022-01-30
Up-to-date tinygo support

sorenisanerd:
Since the original tinygo support was implemented, tinygo has evolved
a lot. It now supports modules, so we don't need to fiddle around with
temporary build directories and explicit "go get" calls, etc.

This patch also changes the default docker image to
tinygo/tinygo:0.22.0 which now works perfectly well for vugu's needs.


gedw99:
hey @sorenisanerd  @bradleypeabody 

i use tnygo and so can help out if needed... 

Wondering what's he status with this ?

Issue #233, opened 2022-09-01
Panic: index out of range


PotatoesFall:
**Describe the bug**
After running smoothly for quite some time, I have an app that is suddenly crashing. Certain actions will trigger a panic with index out of range. This is happening in the `domrender` package in the `(*instructionList).writeValString`, stack trace is included at the end.

I tried to do some checking but I'm not quite sure I understand the `instructionList` well enough to diagnose what is happening.

**Software Versions**
Vugu version: 0.3.4
Go version: 1.18 and 1.19 (both)
Browser and version: Firefox (version unclear), Brave 1.42.97 which is based on Chromium: 104.0.5112.102

**To Reproduce**
unfortunately, I cannot reproduce this bug in a simple way, it seems to occur only with a very specify set of data.

**Expected behavior**
I expect there not to be a panic.

**Additional Notes**
<details>
<summary>Stack Trace</summary>

```
panic: runtime error: slice bounds out of range [:16387] with capacity 16384 wasm_exec.js:51:14
<empty string> wasm_exec.js:51:14
goroutine 1 [running]: wasm_exec.js:51:14
github.com/vugu/vugu/domrender.(*instructionList).writeValString(...) wasm_exec.js:51:14
    /home/marty/go/pkg/mod/github.com/vugu/vugu@v0.3.4/domrender/renderer-js-instructions.go:559 wasm_exec.js:51:14
github.com/vugu/vugu/domrender.(*instructionList).writeSetAttrNSStr(0x848300, {0x0, 0x0}, {0x8ab00, 0x6}, {0x8e428, 0x11}) wasm_exec.js:51:14
    /home/marty/go/pkg/mod/github.com/vugu/vugu@v0.3.4/domrender/renderer-js-instructions.go:204 +0x42 wasm_exec.js:51:14
github.com/vugu/vugu/domrender.(*JSRenderer).syncElement(0x83d1e0, 0x80a180, 0xb8e9c0, {0x885f40, 0x29, 0x40}) wasm_exec.js:51:14
    /home/marty/go/pkg/mod/github.com/vugu/vugu@v0.3.4/domrender/renderer-js.go:584 +0x98 wasm_exec.js:51:14
github.com/vugu/vugu/domrender.(*JSRenderer).visitSyncElementEtc(0x83d1e0, 0x80a180, 0xb66fc0, 0xb6f930, 0xb8e9c0, {0x885f40, 0x29, 0x40}) wasm_exec.js:51:14
[...]

PR #234, opened 2022-09-23
fix index out of bounds

PotatoesFall:
See https://github.com/vugu/vugu/issues/233


PotatoesFall:
@bradleypeabody  any change you can take a look at this?

PR #252, opened 2023-11-19
Lint code with golangci-lint

Lord-Y:
As said in my last comment in PR https://github.com/vugu/vugu/pull/248, here is the 1st of many series of PR.
This PR goes with https://github.com/vugu/vugu/issues/246 requirements.

I usually remove every unused functions/variables for a cleaner code. I didn't do it this time so I add `//nolint:golint,unused` tag. Tell me if you want me to delete them or not.

I'm not sure what you wanted to do https://github.com/Lord-Y/vugu/blob/057f77f040a6752b26d472deff9d9023b379c82e/tinygo-dev/testpgm.go#L19 with `jr` param.

PR #258, opened 2023-11-27
Install golangci-lint with deb package instead of building it

Lord-Y:
Hello @bradleypeabody @owenwaller ,

I'm just changing the way of installing `golangci-lint` to avoid golang version issues.
It's related to https://github.com/vugu/vugu/pull/252#issuecomment-1826965915 for `go 1.17.x`

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,7 +24,7 @@ jobs:
 
     - name: before_install from Travis
       run: go get golang.org/x/tools/cmd/goimports && go install golang.org/x/tools/cmd/goimports
-    - run: wget -O /tmp/golangci-lint.deb https://golangci/golangci-lint/releases/download/v1.55.2/golangci-lint-1.55.2-linux-amd64.deb ; sudo dpkg -i /tmp/golangci-lint.deb
+    - run: wget -O /tmp/golangci-lint.deb https://github.com/golangci/golangci-lint/releases/download/v1.55.2/golangci-lint-1.55.2-linux-amd64.deb ; sudo dpkg -i /tmp/golangci-lint.deb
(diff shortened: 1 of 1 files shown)


Lord-Y:
@owenwaller According to the github workflow, all testing are running on ubuntu so yes it will work. Tinygo was also installed with a deb package.

Lord-Y:
@owenwaller in the PR 252, we have error with golang 1.17.x related to golanci-lint. Building golangci-lint require golang 1.18 minimum. So instead of building it, install the provided version is better an will always work when using lower versions of golang.

owenwaller:
@Lord-Y Ah ok now I get it.

Sorry I tend to move with the go tool chain, so I never see things like trying to build tools and libraries with a older and unsupported version of go. I don't think we should be doing anything that supports building or testing vugu on an unsupported version of go.

In which case I have an alternative suggestion. Rather than make this change on its own can we revert to the `go install` method, but do the version fixing (if that is also important) in PR #257?

That way `golangci-lint` will only be downloaded, build and installed with a go v1.21.X tool chain which it does support.

Does that work?

.github/workflows/go.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
     - name: before_install from Travis
       run: go get golang.org/x/tools/cmd/goimports && go install golang.org/x/tools/cmd/goimports
-    - run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
```
bradleypeabody, line 27: As a general rule I tend to lock the version number (at least the major version) for any tool in use. Otherwise subtle changes can break the build for no good reason.

bradleypeabody:
See comment, but seems okay to me, merging.

PR #263, opened 2023-12-20
Use a Taskfile to build and test vugu

owenwaller:
Use a Taskfile (see: https://taskfile.dev/) to build the vugu packages, the vugugen and vugufmt command line tools, run all of the tests, and lint the source code.

This simplifies building vugu locally. All that is required now is

$ task all

at the top level of the repo. The Taskfile.yaml is fully documented.

Note: this is an initial version that only supports Linux as it relies on command line utilities that are installed by default on Linux. In addition it assumes that docker, tinygo and the golangci-lint tools are all installed locally. If they are not found then the task will abort. Supporting Windows will require changing the tasks to support the Windows PowerShell equivalents when the task in run on Windows.

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -1,0 +1,133 @@
+# Taskfile for 'vugu'
+# 
+# The Taskfile builds vugu, runs all the tests and lints the source code.
+
+version: '3'
+
+# TODO: Warning this task file assumes Linux is the OS. The Taskfile will need to reworked to strip the Linux only shell cmds
+tasks:
+  # Build everything, run all the tests (using go, tinygo and tinygo in docker) and lint the codebase
+  all:
+    preconditions:
+      - task: check-cmds-exist
+    cmds:
+      - task: clean-vgform
+      - task: clean-wasm-test-suite
+      - task: build
+      - task: test
+      - task: test-wasm
+      - task: lint
+
+
+  # Build the vugu packages and the vugugen and vugufmt commands
+  build:
+    cmds:
+      # note we now use the full package names
+      # we have to run go generate in the vgform package
+      - go generate github.com/vugu/vugu/vgform
+      # build and install the vugu package
+      - go install github.com/vugu/vugu
+      # build and install the vugugen command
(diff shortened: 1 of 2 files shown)


Taskfile.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - task: test-wasm
+      - task: lint
+
+
+  # Build the vugu packages and the vugugen and vugufmt commands
+  build:
+    cmds:
+      # note we now use the full package names
+      # we have to run go generate in the vgform package
+      - go generate github.com/vugu/vugu/vgform
```
bradleypeabody, line 27: I don't think we should be re-go-generating the vgform package as part of the build.  The usual process for go generate is that when you edit a package then you go generate it and then the output just stays committed to the repo (there are various examples of this in the Go standard library).  So unless there's a really good reason to include this, I think we can nuke that go generate line

owenwaller: No we need this, at least of the minute.

Without PR261 this generates the "*_vgen" file that has to exist in the repo before the vgfrom package will build.
With PR262 this generates the "*_gen" in the vgform package, which likewise has to exist.

Without these files (named either way) the build fails - because the `vgform` tests fails. If you happen to have both versions on disk at the same time (as I have had when switching between the PR261 branch and this one) then the build will also fail.

If you want to remove it, then we need to somehow ensure that either the `go generate` has run, or the files are present (and correct) in the repo.

As it's more or less impossible to tell the difference between this situations the `Taskfile` takes the only safe option and regenerates everything. This way we are sure everything necessary is on disk before the `go build` steps. The build of everything in `vugu` shouldn't need any special steps to be taken by the developer before hand.

owenwaller:
I've added a new `generate` task which calls `go generate github.com/vugu/vugu/vgform` - which was part of the `build` task.
And added the `generate` to the `all` task, but not made it a prerequisite of the `build` task.

At some future point the github workflow will check if the newly generated files differ from the ones in the repository. If they do then the build will fail, as this implies that the the newly updated generated files have not been committed by the developer.

Issue #181, opened 2021-01-07
vugugen: add one more "unused" import: log


powerman:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I have to constantly add and remove `import "log"` in each component's `<script type='application/x-go'>` while debugging.

**Describe the solution you'd like**
Have `vugugen` to add it by default into _vgen.go files, just like it already does for fmt, reflect, vjson and js packages.

**Describe alternatives you've considered**
An alternative would be make a habit of debugging using fmt.Println instead of log.Println. While this is perfectly fine in frontend part of project, doing same in backend part usually will be a mistake… so it's better to avoid such a "bad" habits.
Another alternative is have this boilerplate in each and every .vugu file (that's what I do now):
```
<script type='application/x-go'>
import "log"
var _ = log.Println // Ensure import is used.
</script>
```

bradleypeabody:
I think it would be more productive to instead see about running the generated output through `goimports`.  This would solve the problem for this package and others.

owenwaller:
If vugu doesn't do this already, then this seems like a worthwhile change to add to v1.0.

PR #270, opened 2024-04-29
Import `log` package by default into `*_gen.go`

miyamo2:
Hello, @bradleypeabody @owenwaller 

This PR is based on Issues #181
`*_gen.go` files generation will be changed in this PR as follows.
```.diff
package main

import "fmt"
import "reflect"
import "github.com/vugu/vjson"
import "github.com/vugu/vugu"
import js "github.com/vugu/vugu/js"
+ import "log"

/*
omit
*/

// 'fix' unused imports
var _ fmt.Stringer
var _ reflect.Type
var _ vjson.RawMessage
var _ js.Value
+ var _ log.Logger
```

Not related to this PR, but I modified the following file to pass CI(golangci-lint).

- wasm-test-suite/docker/wasm-test-suite-srv.go

Thank you.

### DIFF
--- a/gen/parser-go.go
+++ b/gen/parser-go.go
@@ -257,6 +257,7 @@ func (p *ParserGo) visitOverall(state *parseGoState) error {
 	fmt.Fprintf(&state.goBuf, "import %q\n", "github.com/vugu/vjson")
 	fmt.Fprintf(&state.goBuf, "import %q\n", "github.com/vugu/vugu")
 	fmt.Fprintf(&state.goBuf, "import js %q\n", "github.com/vugu/vugu/js")
+	fmt.Fprintf(&state.goBuf, "import %q\n", "log")
 	fmt.Fprintf(&state.goBuf, "\n")
 
 	// TODO: we use a prefix like "vg" as our namespace; should document that user code should not use that prefix to avoid conflicts
@@ -276,6 +277,7 @@ func (p *ParserGo) visitOverall(state *parseGoState) error {
 	fmt.Fprintf(&state.goBufBottom, "var _ reflect.Type\n")
 	fmt.Fprintf(&state.goBufBottom, "var _ vjson.RawMessage\n")
 	fmt.Fprintf(&state.goBufBottom, "var _ js.Value\n")
+	fmt.Fprintf(&state.goBufBottom, "var _ log.Logger\n")
 	fmt.Fprintf(&state.goBufBottom, "\n")
 
 	// remove document node if present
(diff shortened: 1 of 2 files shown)


owenwaller:
Hi @miyamo2 

Thanks for this. Issue #181 seems to fallen off mine and @bradleypeabody radar, but it is the sort of thing we need to fix.

In principle I don't have a problem with the change to `gen/parser-go.go` as that seems like a sensible approach.

But, the second change to `wasm-test-suite/docker/wasm-test-suite-srv.go` is already on the `master` branch as far as I can tell.

See:

https://github.com/vugu/vugu/blob/1782dc212655d99ff1ffe9147402e62317ba1944/wasm-test-suite/docker/wasm-test-suite-srv.go#L151

Do you want to try syncing with the `master` branch and see if that removes the need for the change to `wasm-test-suite/docker/wasm-test-suite-srv.go`?

@bradleypeabody are you ok importing `log` directly, as this does seem like a common case, Or do you still prefer running the `_gen.go` files through `goimports`? I have seen `goimports` pick up the wrong package before now though.

Thanks

Owen

bradleypeabody:
Yeah I guess it's fine to add this log import change on the basis that we already provide the same sort of convenience for `fmt` and `js` packages.  I'm a little leery of using up more import names by default in exchange for a minor convenience - and I would encourage putting more code in separate .go files rather than in `<script type='application/x-go'>` tags.  But I don't really have a compelling answer for why it's fine to do this with `fmt` and not `log`.  So yeah I think if we limit this PR to just the changes in parser-go.go, it's fine to merge.

(And the goimports thing is probably a separate topic, since it would provide good convenience but at the cost of requiring an additional tool and could also have interesting side effects of possibly selecting the wrong package as mentioned.)

miyamo2:
@owenwaller @bradleypeabody 
> But, the second change to wasm-test-suite/docker/wasm-test-suite-srv.go is already on the master branch as far as I can tell.

It was my mistake, it seems it was an unnecessary change, so I restored them.
Thank you.

owenwaller:
@miyamo2 thank you.

@bradleypeabody do you want o merge this one in this? And agreed with the `goimports` comment. It's managed to pick the wrong `log` package for me once today already.

About `<script type='application/x-go'>`...

Personally (and this is only a personal opinion!) I would remove the ability to use `<script type='application/x-go'>` if I possibly could and force all of the Go code into separate `*.go` files. This is one of the things that really threw me a curve ball when I first started trying to do things with `vugu`. Did the Go code live in the `.vugu` file or not? And if it did how the heck did it all get compiled. As soon as I stopped that, and put everything in a `*.go` file named after the component things became easier to understand. So I'd remove these script tags for that reason, but I don't have a `js` background so this may be normal for anyone who has.

But as the use of the `<script type='application/x-go'>` tags is all over the documentation and the tests, I can't see removing these being possible anymore. I also don't want to break anyone's existing `vugu` projects.

What I think we should do is __strrongly__ discourage their use (mark as depreciated, or emit a warning?), update the documentation and all of the `vugu` code so that vugu itself doesn't do it that way. Ideally for any v1.0 release. 

Owen

bradleypeabody:
Cool.  This one is merged.  And @owenwaller I created this ticket https://github.com/vugu/vugu/issues/271 to follow up on the question of what to do with script tags.

PR #272, opened 2024-05-28
Move to a `mage` based build solution for vugu

owenwaller:
This PR moves the `vugu` build process from a YAML based `taskfile` approach to [`mage`](https://magefile.org/). The build process is now platform agnostic provided the host system has `mage`, `docker` and `goimports` installed. 

The previous build approach relied on the [`task`](https://taskfile.dev/) command and a YAML based `Taskfile.yaml`. This YAML file wrapped shell commands. As a result the YAML `Taskfile` was not portable between Linux, MacOS and Windows. The ony solution would have been to create platform specific `Taskfiles` and call them from a top level platform agnostic `Taskfile`.  This approach is unworkable and a large maintenance overhead for contributors and maintainers.

A `mage` based build changes this. All platform dependant operations are now performed in Go and run anywhere Go does. Any 3rd party tools required by the build process are run via dockerized versions. This results in a platform independent build process.

The `mage` built targets can be listed with
```
cd path/to/vugu/
mage -l
```

The PR contains one significant change.

The `wasm-test-suite` has been completely refactored such that:

1) The `wasm` binary is served via an standard `nginx` container, and not a `vugu` `devserver`, Go based web server. The `devserver` and most of the `devutils` package is now largely redundant.
2) The Go Test functions now use a standard `chromedp` container. These two changes remove the need for the `vugu/wasm-test-suite` image.
3) The `nginx` and `chromedp` containers are networked together via a private `docker` network. This allows the `chromedp` container to request the `wasm` binary from the `nginx` server at a known `hostname`. 
4) The `wasm-test-suite` has also been completely refactored to simplify each of the tests. The process of building the test, and starting any required containers is now handled by `mage` rather than the Go `TestXxx` functions. As a result the `wasm-test-suite` is only built with the standard Go compiler. Support for the `tinygo` compiler suite can easily be added again at a later date if required.

Presently all of the `wasm-test-suite` tests pass but tests `test-012-router` and `test-023-lifecycle-callbacks` are **both** incomplete. Both tests stop at the point immediately before the point where they now fail, post refactoring. Both test cases need to be debugged and completed.

The original `wasm-test-suite` is still available, and still builds and passes. It has been moved to the `legacy-wasm-test-suite` package, and cam be built and exxecuted with `mage testLegacyWasm` or `mage allWithLegacyWasm` targets.

Once the two incomplete tests, above, have been completed, the `legacy-wasm-tes-suite` can also be removed.


bradleypeabody:
Thanks very much for all of the incredible work on this @owenwaller , merging now!

PR #273, opened 2024-05-28
Update the magefile to support building and serving the examples

owenwaller:
This PR updates the magefile so that the `vugu` examples can be build and served using `mage`. 

The examples are served using a local `nginx` container. The Readme file has also been updated to show how to run the examples, as well as how to create new ones The existing `fetch-and-display` example has been updated so that it is now built and served using the new approach.

This PR adds two new targets Example and SingleExample to the magefile. Neither target is run by default.
    
The approach mirrors the approach taken for the wasm-test-suite package. Each example is now a Go module in its own sub-directory. The example is built via mage and served using a local nginx container that is serving from the local examples directory. 

Users can then connect to the example using a URL of the form:
    
http://localhost:8888/<name-of-example-directory>
    
e.g. for the fetch and display example
    
http://localhost:8888/fetch-and-display
    
SingleExample is the same as the Example target except it builds the named example module, instead of all examples. To use it you must specify the module name of the example like so:

```
mage singleExample github.com/vugu/vugu/examples/fetch-and-display
```


bradleypeabody:
Can you please clean up the commit history on this or perhaps replay the relevant changes on a branch where I can see the specific differences?  It looks like it's showing 428 files changed - maybe all of the recent changes you've been doing, hard to tell what's going on.

owenwaller:
Reworked as PR #277 which has only the 3 necessary commits.

The reason this commit is huge, is that it was based of the `PR/mage_test001simple_refactor` branch. The PR was created before that branch was merged in to create v0.4.0. So the PR includes all of the commits on `PR/mage_test001simple_refactor` + the 3 additional commits that refer to the examples.

I've now cherry picked the 3 commits that relate the the example onto the v0.4.0 master branch and pushed the PR branch again. The PR should now hopefully be clear.

owenwaller:
This has been superseded by PR #277

PR #275, opened 2024-05-29
Remove unused and unnecessary <script type="application/x-go"> tags

owenwaller:
Removed the used or unnecessary `<script type="application/x-go">` tags from:

./vgform/select.vugu
./vgform/input.vugu
./vgform/textarea.vugu
./wasm-test-suite/test-011-wire/demo-comp1.vugu
./wasm-test-suite/test-011-wire/demo-comp2.vugu
./wasm-test-suite/test-013-issue-117/create.vugu
./wasm-test-suite/test-017-nesting/final.vugu

Where the `<script>` tag has a body the Go code has been moved into a corresponding `*.go` file.

Any remaining `<script type="application/x-go">` tags in the `wasm-test-suite` are necessary - tests `test-012-router` and `test-020-vgform`


bradleypeabody:
This seems fine, but you should `go generate` again in the vgform package so the latest _gen.go files are there.

The reason that vugugen output of _gen.go files is needed here is so that other people not maintaining vugu can import and use that vgform package.  Otherwise other random projects that just do `import "github.com/vugu/vugu/vgform"` will get a compile error.

owenwaller:
I've added the missing `_gen.go` files. They where out of date due to the merge of PR #270, which I'd only picked up after I;d merged in the 0.4.0 changes into my local master branch.

I think PR #276 can now be ignored given the comment.

The merge is due to the fact that I added the `_gen.go` files to a different branch locally (the merge merges form a remote branch in my fork) and then pushed the result to this PR branch.

bradleypeabody:
@owenwaller Thanks and this seems good to me in terms of moving script tags over to .go files.  On select.vugu, I want to keep those old comments in there.  The vgform package is still very much up in the air and most of the problems noted there have not be addressed, so I don't want to nuke those until we figure out what the actual plan is for handling forms in Vugu.  I will go ahead and merge this PR and then add the comments in that file back in.

PR #276, opened 2024-05-29
Delete the auto-generated files from the vgform package

owenwaller:
Delete the auto-generated files from the vgform package

The files

vgform/input_gen.go
vgform/select_gen.go
vgform/textarea_gen.go

Are generated on each build, so there is not need to keep these in the repo.


bradleypeabody:
Please see #275  and let's discuss as needed.

owenwaller:
PR #275 has been updated. Based on the comment in PR #275 this PR is actually redundant.

PR #277, opened 2024-05-30
[Reworked: PR#273] Update the magefile to support building and serving the examples

owenwaller:
This is a reworked version of PR#273 to minimise the number co commits. As per [this comment](https://github.com/vugu/vugu/pull/273#issuecomment-2138724239).

This PR updates the magefile so that the vugu examples can be build and served using mage.

The examples are served using a local nginx container. The Readme file has also been updated to show how to run the examples, as well as how to create new ones The existing fetch-and-display example has been updated so that it is now built and served using the new approach.

This PR adds two new targets Example and SingleExample to the magefile. Neither target is run by default.

The approach mirrors the approach taken for the wasm-test-suite package. Each example is now a Go module in its own sub-directory. The example is built via mage and served using a local nginx container that is serving from the local examples directory.

Users can then connect to the example using a URL of the form:

http://localhost:8888/

e.g. for the fetch and display example

http://localhost:8888/fetch-and-display

SingleExample is the same as the Example target except it builds the named example module, instead of all examples. To use it you must specify the module name of the example like so:

```
mage singleExample github.com/vugu/vugu/examples/fetch-and-display
```


bradleypeabody:
@owenwaller Awesome and thanks for this.  We'll keep discussing the examples and where we want to go from here but this good for now, merging!

PR #287, opened 2024-06-09
Testing out some build changes

bradleypeabody:


owenwaller:
Hi Brad,

I don't believe this is the fix. 

The repo should contain the `go.mod` as it was before the `go generate` was run. If it did then your additional `go mod tidy` before the `go generate` should do nothing.

It looks (maybe) like the wrong `go.mod` is in the repo. If I ensue the `vjson` is a direct dependency (as it is before the `go generate`) then the build still fails.

See my my commit n my fork:

https://github.com/vugu/vugu/compare/master...owenwaller:vugu:PR/fix_tests_go_dot_mod

The corresponding GtiHub Action run will show you the failure.

So I m not convinced ATM. Also I cannot reproduce the issue locally (deleting the generated files makes and running `go mod tidy` either side of go generate makes no difference).

The panic you are seeing in the build failure above I do see locally.

That's due to a change in the `headless-shell` container that was released 13hrs ago (as I write this) so that's some sort of change/bug on their side. As yet I have no idea what's changed.

PR #289, opened 2024-06-25
pinned to older chrome version

bradleypeabody:
This is to fix the issue with the legacy test suite where it looks like it was erroring due to the remote debugging functionality changing.  To fix this up properly to work with the latest version would require further reworking the guts of the docker setup for this so I instead just pinned it to the last working version.  The plan is to get rid of this entirely in the long run so I think this is fine.

### DIFF
--- a/legacy-wasm-test-suite/docker/Dockerfile
+++ b/legacy-wasm-test-suite/docker/Dockerfile
@@ -1,5 +1,5 @@
 
-FROM chromedp/headless-shell:latest
+FROM chromedp/headless-shell:125.0.6422.142
 
 RUN apt-get update && apt-get install dumb-init
 


owenwaller:
I don't think this the fix we want. We don't need to change the `Dockerfile`.

Using `chromedp/headless-shell:latest` as the base image is fine as the latest version works.

All I needed done was to have the docker image rebuilt and pushed to docker hub. The process of rebuilding the image will then pick up the newer `chromedp/headless-shell:latest` (as that will be a different `latest` from the last time the image was built) pushed to docker hub.

At that point I think things will be ok again.

Agreed this will be removed at some point RSN.

bradleypeabody:
@owenwaller the problem is the changes I made don't work with `chromedp/headless-shell:latest`.  I will go ahead and merge this as is and let's just see what the status is after that.

PR #291, opened 2024-07-11
Make the building of the tests consistent with the vgform build process

owenwaller:
By removing the generate.go files and calling vugugen directly as part of the magefile build process we have created an incnsistency in the build process.

The vgform package has always had the generated "_gen.go" files commited to the the repo. This is required to make the vgfrom package "go getable" and becuase the tests - especilly the linting - relies on these files being present.

But this was not the case for the tests in the wasm-test-suite and the examples packages.

This commit brings the wasm-test-suite and examples packages into line with the approach used for the vgform package.

For each test in the wasm-test-suite the "*_gen.go" files are now committed to the repo along with the go.mod and go.sum after both vugugen and go mod tidy have bene run. The local .gitignore files in each test have also been updated so that the "*._gen.go" files are no longer ignored.

The same change has also been made to the examples/fetch-and-display package.

To support these changes the magefile has been updated. There is now a new "clean" target. This only removes any generated *.wasm wasm files. It is called as a prerequisite of the "Build" target.

The existing "CleanAutoGeneratedFiles" target is still present, should it be needed, but it will now only remove the "*_gen.go" files created by vugugen. It is no longer a prerequisite of the "Build" or any other target.


bradleypeabody:
Works for me. Merging.

PR #306, opened 2024-08-28
An initial, simple port of the vue conditionals and loops example

owenwaller:
An inital port of the Vue conditionals and loops example taken from: https://vuejs.org/examples/#conditionals-and-loops

To run this example can be run using the magefile with:

mage singleExample github.com/vugu/vugu/examples/vg-if

and then browsing to:

http://localhost:8889/vg-if/

Note: the implmentation uses the Go standard library slices package, available since Go v1.21.


bradleypeabody:
Cool! The example itself seems fine to me.

The only thing I would like to address - and we can discuss this directly in Slack if we need to hash it out more, is that I think it would be better if the commands required to build this example were at least documented somewhere in this folder.  So for example, in README.md could this include the few lines necessary to build the example and run it in an nginx container?  This way someone could easily copy this to their own project and drop those commands into their build system of choice and boom they are up and running.

owenwaller:
Ok I've tidied the code up, added some comments and added a Readme.

The Readme details how to build and run the example using the `magefile`. Since you have to clone the `vugu` repo to get the example source code this is by far the simplest approach.

What I propose - but not in this PR is a separate top level Readme at the root of the  `examples` package that details how to run an example without the using `mage`.

As this is (almost but not quite) the same as the existing `wiki` page. I propose to update the wiki and then either link to the wiki from the Readme (preferred) or duplicate the content - which might be easier if browsing the repo on Github.

bradleypeabody:
Please fix the duplicate readme.md file (there's two in there with different cases).  Otherwise seems good.  I still would prefer to see the exact commands for each example in the individual example readme, but I won't die over it at this point.

owenwaller:
The duplicate Readme file has been resolved in #[1f31225](https://github.com/vugu/vugu/pull/306/commits/1f31225fae93601d9d05a08a8078d4a2065f88d7)

PR #317, opened 2024-10-15
Fix: add `wasm_exec.js` to the local git ignore file

owenwaller:
wasm_exec.js shoudl always be copied from the GOROOT as part of the build porcess as it has to match the Go compiler version. So we want to ignore any local copies.


owenwaller:
See also PR #345. That PR does not include the `vg-if` example - as that has already been done via this PR.

owenwaller:
Closing as this is no longer required. It's been overtaken by PR #345

Issue #266, opened 2024-01-15
js Set can not set Function


CJBigsharp:
**Describe the bug**
js Set can not set a Function ,it will be throw panic: ValueOf: invalid value

**Software Versions**
Vugu version: 
Go version: 
Browser and version:

**To Reproduce**
Please describe how to reproduce the bug you encountered.

If possible, please include a concise code snippet which reproduces the issue (and excludes other unnecessary functionality, for clarity).

**Expected behavior**
check the js.Set, and add a judge for funtion

**Additional Notes**
Anything else...

CJBigsharp:
func (v Value) Set(p string, x interface{}) {
	// judge x is FuncOf
	if f, ok := x.(Func); ok {
		sjs.Value(v).Set(p, f.f)
	} else {
		sjs.Value(v).Set(p, x)
	}

}

owenwaller:
Hi @CJBigsharp 

Thanks for this, but can you please provide a complete working example of what you are trying to do, and _why_? 
Can you also tell us what version of Go and vugu you have been using, and how you compiled the code? We also need to see the full panic trace.

I suspect, this may be noting to do with vugu, if and only if `sjs` is a reference to the [`syscall/js` package](https://pkg.go.dev/syscall/js). In this case you would expect `Set` to panic when `v` is not a valid javascript object. This is the [documented behaviour of `Set`](https://pkg.go.dev/syscall/js#Value.Set).

If this is the case then my _guess_ is that `sjs.Value(v)` is not returning you a javascript object, which causes `Set` to panic.

But without more context that is just a _guess_.

CJBigsharp:
> Hi @CJBigsharp
> 
> Thanks for this, but can you please provide a complete working example of what you are trying to do, and _why_? Can you also tell us what version of Go and vugu you have been using, and how you compiled the code? We also need to see the full panic trace.
> 
> I suspect, this may be noting to do with vugu, if and only if `sjs` is a reference to the [`syscall/js` package](https://pkg.go.dev/syscall/js). In this case you would expect `Set` to panic when `v` is not a valid javascript object. This is the [documented behaviour of `Set`](https://pkg.go.dev/syscall/js#Value.Set).
> 
> If this is the case then my _guess_ is that `sjs.Value(v)` is not returning you a javascript object, which causes `Set` to panic.
> 
> But without more context that is just a _guess_.

sjs:
func (v Value) Set(p string, x interface{}) {
	if vType := v.Type(); !vType.isObject() {
[...]

Issue #330, opened 2025-01-23
All `vugu` generated files are missing `//go:build js && wasm` build tags


owenwaller:
**Describe the bug**

As part of the investigation into issue #328 I was attempting to update the new `test-025-valueof` example to this:

```go
package main

import (
	"fmt"
	"time"

	sysjs "syscall/js" // <-- new

	js "github.com/vugu/vugu/js"
)

type Root struct {
}

// Make a vugu.js.ValueOf call on an object derived from a JS global object - in this case the global "Date" object.
// Returns the string "Ok" or a panic message
// See Issue #328 https://github.com/vugu/vugu/issues/328
// vugu/js version
func (c *Root) VuguJsValueOf() (s string) {
	// We have to be a little clever here
	// We need to recover from the panic but change the value we return.
	// The simplest way to achieve this is with a named return value.
	// By setting the named return value ot the panic string we will overwrite any previous
	// value (in this case "Ok") that we would otherwise return.
	defer func() {
		if r := recover(); r != nil {
			s = fmt.Sprintf("%s", r) // set the names return value to the panic string
		}
	}()
	c.vuguJsPanicingFunc()
	return "Ok"
}

// Syscall/js version of VuguJsValueOf
func (c *Root) SyscallJsValueOf() (s string) { // <-- new
	defer func() {
		if r := recover(); r != nil {
			s = fmt.Sprintf("%s", r) // set the names return value to the panic string
		}
	}()
	c.syscallJsPanicingFunc()
	return "Ok"
}

// panicingFunc makes the vugu.js.ValueOf call that rill result in a panic
// vugu/js version
func (c *Root) vuguJsPanicingFunc() {
	date := js.Global().Get("Date") // NOTE: use of double quotes not single quotes as in Issue #328 https://github.com/vugu/vugu/issues/328
	timeEndValue := date.New(time.Now().UnixMilli())
	// This line with panic with:
	//
	// ValueOf: invalid value
	//
	// The underlying implementation of js.Value() calls syscall.js.Value.
	// The underlying type of the interface{} that is passed to syscall.js.Value is a vugu.js.Value
	// because a vugu.js.Value is the return type of the js.Global.Get(...) call above.
	// But because a vugu.js.Value is not a known type inside syscall.js.ValueOf the result is a panic from syscall.js.ValueOf.
	// The internal implementation of syscall.js.ValueOf consists of a type switch statement i.e. `switch x := x.(type)`
[...]

PR #331, opened 2025-01-24
Fix and rework vugu/js

### LINKED ISSUE #328: Nested value in array/object doesn't work
**Describe the bug**
If you need a more complex type nested in an array or object the js wrapper library doesn't properly cooperate with syscall/js. The problem is that the ValueOf function in syscall/js checks whether the passed argument is a Value object. However, since the object comes from the vugu js implementation it doesn't match syscall/js.Value.

The simplest way to reproduce the issue is to call js.ValueOf(someValueObject).

As a workaround I just used syscall/js directly but that obviously isn't ideal. I initially expected the JSValue() function to help, but that doesn't appear to do anything.

**Software Versions**
Vugu version: v0.4.0
Go version: v1.23.3
Browser and version: Chrome 131.0.6778.264 

**To Reproduce**
The running the following code in a wasm JS context causes a panic:
```go
import (
    "time"

    js "github.com/vugu/vugu/js"
)

type Root struct {
}

func (c *Root) Init (ctx vugu.InitCtx) {
    date := js.Global().Get('Date')
    timeEndValue := date.New(time.Now().UnixMilli())
    js.ValueOf(timeEndValue)
}
```

**Expected behavior**
ValueOf should pass the object without problems.

**Additional Notes**
 A workable solution would be if JSValue handled the cast for you if necessary depending on the compilation context.

Dadido3:
`vugu/js` is now in sync with `syscall/js`.
The following changes and fixes were applied:

- Original comments were added, so most editors will show useful info, no matter if placeholder or alias types are used.
- Moved `fixArgsToSjs` into `helpers.go` and renamed it to `convertAnyToSJS`
- Fix missing conversion of return values in `FuncOf`
- Use go type aliasing where possible
- Add missing `Equal` method to `Value`
- Add missing `IsNaN` method to `Value`
- Fix slice and map handling of `ValueOf`
- Fix missing conversion in `Set` method of `Value`
- Fix missing conversion in `SetIndex` method of `Value`
- Make nonjs `Value` uncomparable
- Add missing nonjs placeholder types and functions
- Make behavior of the placeholder types and functions more consistent with the description in `doc.go`

This makes #205 and #207 obsolete. And should fix #328 and #266.

I fully tested it by using the original `syscall/js` tests. Also, it may be good to move some/all of the tests to the wasm test suite, and add some more tests related to slices and maps containing `vugu/js` Value objects...

@owenwaller I haven't included any wasm unit test, as you are already working on one.

### DIFF
--- a/js/helpers.go
+++ b/js/helpers.go
@@ -1,0 +1,49 @@
+//go:build js && wasm
+
+package js
+
+import sjs "syscall/js"
+
+// convertAnyToSJS converts any aliased value into its syscall/js counterpart.
+func convertAnyToSJS(arg any) any {
+	switch arg := arg.(type) {
+	case Value:
+		return sjs.Value(arg)
+	case Func:
+		return arg.f
+	}
+
+	return arg
+}
+
+// convertSliceToSJS takes a []any and converts every aliased value into its syscall/js counterpart.
+//
+// This is a bit dirty, as it may modify the given slice.
+func convertSliceToSJS(args []any) []any {
+	for i, arg := range args {
+		switch arg := arg.(type) {
+		case Value:
+			args[i] = sjs.Value(arg)
+		case Func:
+			args[i] = arg.f
+		}
+	}
+
+	return args
+}
+
+// convertMapToSJS takes a map[T]any and converts every aliased value into its syscall/js counterpart.
+//
+// This is a bit dirty, as it may modify the given map.
+func convertMapToSJS[T comparable](args map[T]any) map[T]any {
+	for k, arg := range args {
+		switch arg := arg.(type) {
+		case Value:
+			args[k] = sjs.Value(arg)
+		case Func:
+			args[k] = arg.f
+		}
+	}
+
+	return args
+}
--- a/js/impl-js.go
+++ b/js/impl-js.go
@@ -1,4 +1,4 @@
-// +build js
+//go:build js && wasm
 
 package js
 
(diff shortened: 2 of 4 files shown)


owenwaller:
@Dadido3 

Ah.. it seems I need to update the Contributing guide, as you won't actually have known this....but...

Is there any chance you can recreate this PR for me?

What I need yo do it is push this change to a branch in your fork called `PR/syscall-js-fix` the `PR/` bit is important.

Any PR that we get with a branch name of that form will (or should!) automatically run the fill Github Action pipeline to tell us if the build and all the tests pass. We can then merge things in.

If you want to sanity check locally before you push anything the magic command is:

`mage allGitHubAction`

That could take anything up to 35mins to run. But if it passes locally there is a very high chance it will also pass when the the PR is created.

@bradleypeabody and I can then take a look at things.

As I have said in Issue #330 I don't think this is a solution but it's not my sole call.

One thing that is interesting is that you managed to run teh Go `syscall/js` test suite. Do you want to tell me how you did that? The comment at the top of the `impl-js-test.go` file implies we need `node.js`.

But for me a better solution would be to use the [wasmtime](https://docs.wasmtime.dev/) tool from the [bytecode alliance](https://bytecodealliance.org/) wo as far as I know are the folks who look after the wasm standard.

The reason I am suggesting this is that the wasmtime tool has Go bindings, so we can call wasm code directly from a Go test and have the wasm test run (inside its on wsam runtime, inside the Go test process).

There is an example [here](https://docs.wasmtime.dev/lang-go.html). I am wondering if this could also opens up a new testing approach for `vugu`

Dadido3:
Alright, got a new PR up, gonna close this in favor of the newer.

> As I have said in Issue #330 I don't think this is a solution but it's not my sole call.

Yes, but it's definitely better than the current `vugu/js`, so i don't see a reason not to pull it for now to have the issues fixed immediately until a better solution is implemented.

> One thing that is interesting is that you managed to run teh Go syscall/js test suite. Do you want to tell me how you did that? The comment at the top of the impl-js-test.go file implies we need node.js.

The original `js_test.go` mentioned how this works in principle, i just had to fix the path to the `go_js_wasm_exec` bash file. It worked surprisingly well and quick. But unfortunately it only works on linux, and as it uses node.js there is no DOM, so you can't run the vugu tests against it unfortunately.

> But for me a better solution would be to use the [wasmtime](https://docs.wasmtime.dev/) tool from the [bytecode alliance](https://bytecodealliance.org/) wo as far as I know are the folks who look after the wasm standard.
> 
> The reason I am suggesting this is that the wasmtime tool has Go bindings, so we can call wasm code directly from a Go test and have the wasm test run (inside its on wsam runtime, inside the Go test process).

I expect that to have the same problem as the node.js solution. There may be no DOM, and in case of the WASM runtime probably not even a JavaScript engine that would be needed to run any vugu tests. But that's just a guess without a deeper dive into it.

PR #332, opened 2025-01-25
Fix and rework vugu/js

### LINKED ISSUE #328: Nested value in array/object doesn't work

Dadido3:
`vugu/js` is now in sync with `syscall/js`.
The changes compared to the current version of `vugu/js` are:

- Add comments from `syscall/js`, so most editors will show useful info, no matter if placeholder or alias types are used.
- Move `fixArgsToSjs` into `helpers.go` and rename it to `convertAnyToSJS`
- Fix missing conversion of return values in `FuncOf`
- Use go type aliasing where possible
- Add missing `Equal` method to `Value`
- Add missing `IsNaN` method to `Value`
- Fix slice and map handling of `ValueOf`
- Fix missing conversion in `Set` method of `Value`
- Fix missing conversion in `SetIndex` method of `Value`
- Make nonjs `Value` uncomparable
- Add missing nonjs placeholder types and functions
- Make behavior of the placeholder types and functions more consistent with the description in `doc.go`

This makes #205 and #207 obsolete. And should fix #328 and #266.

I fully tested it by using the original `syscall/js` tests. Also, it may be good to move some/all of the tests to the wasm test suite, and add some more tests related to slices and maps containing `vugu/js` Value objects...

Recreation of PR #331 due to wrong branch naming.

### DIFF
--- a/js/helpers.go
+++ b/js/helpers.go
@@ -1,0 +1,49 @@
+//go:build js && wasm
+
+package js
+
+import sjs "syscall/js"
+
+// convertAnyToSJS converts any aliased value into its syscall/js counterpart.
+func convertAnyToSJS(arg any) any {
+	switch arg := arg.(type) {
+	case Value:
+		return sjs.Value(arg)
+	case Func:
+		return arg.f
+	}
+
+	return arg
+}
+
+// convertSliceToSJS takes a []any and converts every aliased value into its syscall/js counterpart.
+//
+// This is a bit dirty, as it may modify the given slice.
+func convertSliceToSJS(args []any) []any {
+	for i, arg := range args {
+		switch arg := arg.(type) {
+		case Value:
+			args[i] = sjs.Value(arg)
+		case Func:
+			args[i] = arg.f
+		}
+	}
+
+	return args
+}
+
+// convertMapToSJS takes a map[T]any and converts every aliased value into its syscall/js counterpart.
+//
+// This is a bit dirty, as it may modify the given map.
+func convertMapToSJS[T comparable](args map[T]any) map[T]any {
+	for k, arg := range args {
+		switch arg := arg.(type) {
+		case Value:
+			args[k] = sjs.Value(arg)
+		case Func:
+			args[k] = arg.f
+		}
+	}
+
+	return args
+}
--- a/js/impl-js.go
+++ b/js/impl-js.go
@@ -1,4 +1,4 @@
-// +build js
+//go:build js && wasm
 
 package js
 
@@ -24,58 +24,173 @@ package js
 // go away, can't really tell; my crystal ball is broken.
 // See: https://hacks.mozilla.org/2019/03/standardizing-wasi-a-webassembly-system-interface/
 
-import sjs "syscall/js"
+import (
+	sjs "syscall/js"
+)
 
-// import "unsafe"
+// ######################
+// # syscall/js func.go #
+// ######################
 
-// Error alias to syscall/js
-type Error struct {
-	// Value is the underlying JavaScript error value.
-	Value
-}
-
-// Error alias to syscall/js
-func (e Error) Error() string {
-	return sjs.Error{Value: sjs.Value(e.Value)}.Error()
-	// return (*(*sjs.Error)(unsafe.Pointer(&e))).Error()
-}
-
-// Func alias to syscall/js
+// Func is a wrapped Go function to be called by JavaScript.
+//
+// This is a syscall/js alias.
 type Func struct {
-	Value
-	f sjs.Func // proxy for this func from syscall/js
+	Value          // A copy of the syscall/js Value inside of f, casted into Value.
+	f     sjs.Func // The original syscall/js Func object.
 }
 
-// FuncOf alias to syscall/js
-func FuncOf(fn func(Value, []Value) interface{}) Func {
-
-	fn2 := func(this sjs.Value, args []sjs.Value) interface{} {
-		args2 := make([]Value, len(args))
-		for i := range args {
-			args2[i] = Value(args[i])
+// FuncOf returns a function to be used by JavaScript.
+//
+// The Go function fn is called with the value of JavaScript's "this" keyword and the
+// arguments of the invocation. The return value of the invocation is
+// the result of the Go function mapped back to JavaScript according to ValueOf.
+//
+// Invoking the wrapped Go function from JavaScript will
+// pause the event loop and spawn a new goroutine.
+// Other wrapped functions which are triggered during a call from Go to JavaScript
+// get executed on the same goroutine.
+//
+// As a consequence, if one wrapped function blocks, JavaScript's event loop
+// is blocked until that function returns. Hence, calling any async JavaScript
+// API, which requires the event loop, like fetch (http.Client), will cause an
+// immediate deadlock. Therefore a blocking function should explicitly start a
+// new goroutine.
+//
+// Func.Release must be called to free up resources when the function will not be invoked any more.
+//
+// This is a syscall/js alias.
+func FuncOf(fn func(this Value, args []Value) any) Func {
+	fnConv := func(this sjs.Value, args []sjs.Value) any {
+		argsConv := make([]Value, len(args))
+		for i, arg := range args {
+			argsConv[i] = Value(arg)
 		}
-		return fn(Value(this), args2)
+		return convertAnyToSJS(fn(Value(this), argsConv))
 	}
 
-	f := sjs.FuncOf(fn2)
+	f := sjs.FuncOf(fnConv)
 
-	ret := Func{
+	return Func{
 		Value: Value(f.Value),
 		f:     f,
 	}
-
-	return ret
 }
 
-// Release alias to syscall/js
+// Release frees up resources allocated for the function.
+// The function must not be invoked after calling Release.
+// It is allowed to call Release while the function is still running.
+//
+// This is a syscall/js alias.
 func (c Func) Release() {
-	// return (*(*sjs.Func)(unsafe.Pointer(&c))).Release()
 	c.f.Release()
 }
 
-// Type alias to syscall/js
-type Type int
+// ####################
+// # syscall/js js.go #
+// ####################
 
+// Value represents a JavaScript value. The zero value is the JavaScript value "undefined".
+// Values can be checked for equality with the Equal method.
+//
+// This is a syscall/js alias.
+type Value sjs.Value
+
+// Error wraps a JavaScript error.
+//
+// This is a syscall/js alias.
+type Error = sjs.Error
+
+// Error implements the error interface.
+//
+// This is a syscall/js alias.
+// func (e Error) Error() string
+
+// Equal reports whether v and w are equal according to JavaScript's === operator.
+//
+// This is a syscall/js alias.
+func (v Value) Equal(w Value) bool {
+	return sjs.Value(v).Equal(sjs.Value(w))
+}
+
+// Undefined returns the JavaScript value "undefined".
+//
+// This is a syscall/js alias.
+func Undefined() Value {
+	return Value(sjs.Undefined())
+}
+
+// IsUndefined reports whether v is the JavaScript value "undefined".
+//
+// This is a syscall/js alias.
+func (v Value) IsUndefined() bool {
+	return sjs.Value(v).IsUndefined()
+}
+
+// Null returns the JavaScript value "null".
+//
+// This is a syscall/js alias.
+func Null() Value {
+	return Value(sjs.Null())
+}
+
+// Null returns the JavaScript value "null".
+//
+// This is a syscall/js alias.
+func (v Value) IsNull() bool {
+	return sjs.Value(v).IsNull()
+}
+
+// IsNaN reports whether v is the JavaScript value "NaN".
+//
+// This is a syscall/js alias.
+func (v Value) IsNaN() bool {
+	return sjs.Value(v).IsNaN()
+}
+
+// Global returns the JavaScript global object, usually "window" or "global".
+//
+// This is a syscall/js alias.
+func Global() Value {
+	return Value(sjs.Global())
+}
+
+// ValueOf returns x as a JavaScript value:
+//
+//	| Go                     | JavaScript             |
+//	| ---------------------- | ---------------------- |
+//	| js.Value               | [its value]            |
+//	| js.Func                | function               |
+//	| nil                    | null                   |
+//	| bool                   | boolean                |
+//	| integers and floats    | number                 |
+//	| string                 | string                 |
+//	| []interface{}          | new array              |
+//	| map[string]interface{} | new object             |
+//
+// Panics if x is not one of the expected types.
+//
+// This is a syscall/js alias.
+func ValueOf(x any) Value {
+	switch x := x.(type) {
+	case []any:
+		return Value(sjs.ValueOf(convertSliceToSJS(x)))
+
+	case map[string]any:
+		return Value(sjs.ValueOf(convertMapToSJS(x)))
+
+	default:
+		return Value(sjs.ValueOf(convertAnyToSJS(x)))
+
+	}
+}
+
+// Type represents the JavaScript type of a Value.
+//
+// This is a syscall/js alias.
+type Type = sjs.Type
+
+// This is a syscall/js alias.
 const (
 	TypeUndefined = Type(sjs.TypeUndefined)
 	TypeNull      = Type(sjs.TypeNull)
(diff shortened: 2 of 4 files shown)


owenwaller:
CC'd @RobinGeuze

@Dadido3 

I agree that this version of `vugu/js` is an improvement over what exists currently.

Can you confirm is this is a fix for Issue #328?

owenwaller:
@Dadido3 

Why isn't the file `impl-js.go` just an example of the object adapter pattern?

In other words given that you have this:

```go
// This is a syscall/js alias.
type Value sjs.Value
```
and
```go
type Func struct {
	Value          // A copy of the syscall/js Value inside of f, casted into Value.
	f     sjs.Func // The original syscall/js Func object.
}
```
why doesn't this (for example)

```go
func FuncOf(fn func(this Value, args []Value) any) Func {
	fnConv := func(this sjs.Value, args []sjs.Value) any {
		argsConv := make([]Value, len(args))
		for i, arg := range args {
			argsConv[i] = Value(arg)
		}
		return convertAnyToSJS(fn(Value(this), argsConv))
	}

	f := sjs.FuncOf(fnConv)

	return Func{
		Value: Value(f.Value),
		f:     f,
	}
}
```

collapse to:

```go
func FuncOf(fn func(this Value, args []Value) any) Func {
	f := sjs.FuncOf(fn)
	return Func{
		Value: Value(f.Value),
		f:     f,
	}
}
```

The same applies to the types so for me `Func` is this:

```go
type Func struct {
    sjs.Func // The original syscall/js Func object.
}

```

*but* and even better solution to this is might be to use a new type alias. See the [Go blog](https://go.dev/blog/alias-names).

So in `vugu/js` we would have this:

```go
//go:build js && wasm

package js
import (
	sjs "syscall/js"
)

type Value = sjs.Value
type Func = sjs.Func
...
```

This might be the ideal solution as what we are really trying to do here is to akin to an incremental refactor where we want one type in a new package (`vugu/js/Value`) to refer to another type in a different package `syscall/js/Value`.

owenwaller:
@Dadido3 

ATM I am becoming more and more convinced that `vugu/js` is an unworkable solution. 

If `vugu/js` is not a straight forward adapter pattern (possibly with Go type aliasing) which is what you seem to be saying then that rules out a tool that would automatically generate a new wrapper with every new `syscall/js` release. Which leads to a burden on `vugu` maintainers.

As someone who has written a __lot__ of code that has been cross compiled to other targets it is highly unusual to wrap a `syscall` package. You would work at a higher level of abstraction and then at build time pull in the required low level package directly. This is really what the build tags would do. This would also imply that `vugu/js` is not at the correct level of abstraction - a higher one is needed. Also a higher level of abstraction might also lead to a simpler mock/stub implementation of that higher level abstraction for testing on the native platform - i.e. `!js && !wasm`. 

For the minute the options seem to be:

a) maintaining the version of `vugu/js` that maps onto `syscall/js`. This PR looks like an improvement on this point over the current `vugu/js` package.
b) maintaining the mock version of `vugu/js` for other targets
c) And even if we do (a) and (b) as a pair which we seem to have to do, a user still can't make a call to `syscall/js` directly. This means using `vugu` is preventing a developer for using a legitimate Go package, if using `syscall/js` causes the build of their application (or the `vugu` tests) to fail. If the developer pulls in a 3rd party Go package that uses `syscall/js` somewhere internally then they can't use that 3rd party package either. Importantly, the calls to `sycall/js` cannot be replaced with calls to `vugu/js` in this case.

Now this might only be an issue in test code if (as is currently the case) the `vugu` code is built natively (against a stub `vugu/js`), before the test code is built. But, this is how we test currently within the `vugu` project. 

We have no idea how developers who are using `vugu` to build an application are testing (or even if they are testing) their `vugu` components. But I can see a lot of Github issues in the future along the lines of "component XXX is undefined" when people try to build and test. These will all come down to a reference to `syscall/js` somewhere - which can't be replaced by a `vugu/js` reference (think in a 3rd party package).

Just for reference according to [pkg.go.dev](https://pkg.go.dev/syscall/js?tab=importedby) there are about 3000 imports of `syscall/js` at this point.

So (c) is the real problem, and this PR won't fix that, even if it helps with (a) and (b)

I don't like (a) and (b) but I cannot live with (c) still being in place. I don't think ATM telling developers they can't use `syscall/js` but have to always use `vugu/js` is a solution. There is no way to use tooling to enforce that restriction.

Where as if build tags are a solution, then I would hope that the build would fail in a way that makes it clear build tags need to be added (or removed I suppose).

At this point I am considering ripping every reference to `vugu/js` out, and replacing that with `syscall/js` and then rebuilding everything to see what breaks (I suspect this will break some of the unit tests as well as the static renderer package as a minimum). 

It would take some time to do, but it might help us gauge the problem a bit better.

Dadido3:
> a user still can't make a call to `syscall/js` directly

A user can always use `syscall/js`. It's also possible to mix the two packages and pass a `syscall/js` value into `vugu/js` functions.

---

For example to have code that is specific to browsers and native platforms:

`some-component.go`:
```go
type SomeComponent struct {}
```

`some-component_native.go`:
```go
//go:build !js || !wasm

func (c *SomeComponent) handleButton(event input.ClickEvent) {} // Do nothing.
```

`some-component_js.go`:
```go
//go:build js || wasm

func (c *SomeComponent) handleButton(event input.ClickEvent) {
    js.Global()... // Do something with `syscall/js`.
}
```

---

Or if you don't care about the native part:

`some-component.go`:
```go
//go:build js || wasm

type SomeComponent struct {}

func (c *SomeComponent) handleButton(event input.ClickEvent) {
    js.Global()... // Do something with `syscall/js`.
}
```

---

And the alternative to "Give all generated files build tags" would be this:

`vugugen` could mirror the build tags from the component logic files (E.g. `some-component.go`) when generating code. This is backwards compatible and gives the user total freedom:

- Have all files with `//go:build js && wasm`, even generated ones
- Mixed use of `//go:build js && wasm` and `//go:build !js || !wasm`
- No use of build tags at all
- Use `vugu/js` or `syscall/js` however you want
- Mixed use of `vugu/js` and `syscall/js`
- ...

---

> At this point I am considering ripping every reference to `vugu/js` out, and replacing that with `syscall/js` and then rebuilding everything to see what breaks (I suspect this will break some of the unit tests as well as the static renderer package as a minimum).

I would be interested to see the result of this.

owenwaller:
@Dadido3 

> A user can already use syscall/js. It's also possible to mix the two packages and pass a syscall/js value into vugu/js functions.

This is false.

I have just merged this PR into the [PR/wasm-test-025-valueof-with-syscall-js branch](https://github.com/owenwaller/vugu/tree/PR/wasm-test-025-valueof-with-syscall-js) in my fork. This branch contains an additional test case that explicitly makes a call to `syscall/js` form a `root.go` file.

I am seeing the same failure mode as I did before the merge. To be clear I _expect_ (and understand the failure mode) of my branch `PR/wasm-test-025-valueof-with-syscall-js`. This is Issue #330.

A simple `mage testSingleWasmTest github.com/vugu/vugu/wasm-test-suite/test-025-valueof` will reproduce the failure mode.

```
$ mage testSingleWasmTest github.com/vugu/vugu/wasm-test-suite/test-025-valueof
    ...
    package github.com/vugu/vugu/wasm-test-suite/test-025-valueof
            imports syscall/js: build constraints exclude all Go files in /home/owen/sdk/go1.23.5/src/syscall/js
    ...
    Error: running "go test -v github.com/vugu/vugu/wasm-test-suite/test-025-valueof" failed with exit code 1
```

What I was expecting was that this PR would resolve this and allow the test to pass. This is not the case.

Dadido3:
> What I was expecting was that this PR would resolve this and allow the test to pass. This is not the case.

This PR (#332) never did intend to change this behavior, and it was never communicated this way. All it does is to bring `vugu/js` up to date and fix some issues so that the package does what it was designed to do. It doesn't change how you can or should use the package. Also, it doesn't have any effects on where you can use `syscall/js`.

> > A user can already use syscall/js. It's also possible to mix the two packages and pass a syscall/js value into vugu/js functions.
> 
> This is false.

Check [the comment above](332#issuecomment-2620170668) to see how get a package run on both, native and browser platforms. Or use any of the other methods discussed in #330.

owenwaller:
@Dadido3 

Your [comment](https://github.com/vugu/vugu/pull/332#issuecomment-2620170668) actually agrees with the build tags argument I have been making in Issue #330.

The only difference, and this is a suggestion I am inclined to agree with, is to put the `component` type struct in a separate `*.go` file. This file has to build on all platforms (i.e. native and `GOOS=js GOARCH=wasm`). But split the implementation out into separate files, one for each target platform (i.e. native and `GOOS=js GOARCH=wasm`). You seem to be proposing that this split is enforced through build tags. This is more of less what Issue #330 proposes.

The proposed file split could also be the higher level of abstraction than a wrapper over a `syscall` package, i.e. `vugu/js` that I proposed [here](https://github.com/vugu/vugu/pull/332#issuecomment-2619829520)

>  it is highly unusual to wrap a syscall package. You would work at a higher level of abstraction and then at build time pull in the required low level package directly

Now...

Because of the way `vugu` names files, and the significance of the file name structure, I suspect (but have not tried) that file names like this `component_js_wasm.go` won't be matched to a corresponding `component.go` or `component.vugu` file. I think to support this we would have to change `vugugen`. I have not investigated such a change.

So, for the minute we might need to use the build tags rather than the equivalent file extensions. 

I have checked that you can have a file set named like this `component.vugu`, `component.go`, `implComponent.go` and that `vugugen` will still generate a `component_gen.go` provided everything is in the same package. `implComponent.go` was deliberately chosen so that it did not start with the word `component`.

So, it now seems we both agree we need build tags and that the solution works, after clarification of which files need build tags.

If a developer is creating an application with `vugu` needs to make use of `syscall/js` directly then they need to put the appropriate build tag in the `*.go` file that implements that component and provide a native mock implementation of the same component.

This seems to break my option (c) constraint [in this comment](https://github.com/vugu/vugu/pull/332#issuecomment-2619829520)

This leaves three open questions:

a) What does `vugugen` do, and should there be build tags in the `*_gen.go` files?
b) What should the internal `vugu` tests do?
c) What do we need to get this PR merged?

### `vugugen`

So as far as I can tell ATM the `component_gen.go` file that `vugugen` currently uses will just work. No build tags are required. The `*.gen.go` files never use either `vugu/js` or `syscall/js`. 

The only reason that `vugu/js` is included as a blank variable of type `js.Value` is to support embedding Go code inside `<script type="application/x-go">` tags in a `*.vugu` file. *Personal note:* I don't like this and can think of ways to eliminate it (along with `<script type="application/x-go">` tags. Go code should be in a `*.go` file. But this is supported ATM.

I don't know if this requirement will change in the future, or if we will have to insert build tags into the `*_gen.go` files or perhaps generate a number of different `*_gen.go` files, one for each target platform.

### Internal `vugu` tests

So this goes all the way back to my [comment](https://github.com/vugu/vugu/issues/330#issuecomment-2612839711) in Issue #330. 

I think we would be much better of for the `vugu` project (and the `vugu` project only) if we were more explicit about which target we build a test for (native) and the wasm file (`GOOS=js GOARCH=wasm`)  we want to serve to the test .

If this was made explicit then failures like this:

```
$ mage testSingleWasmTest github.com/vugu/vugu/wasm-test-suite/test-025-valueof
    ...
    package github.com/vugu/vugu/wasm-test-suite/test-025-valueof
            imports syscall/js: build constraints exclude all Go files in /home/owen/sdk/go1.23.5/src/syscall/js
    ...
    Error: running "go test -v github.com/vugu/vugu/wasm-test-suite/test-025-valueof" failed with exit code 1
```

would not occur.

To do this, it needs a change in the file structure of the `vugu` project, and changes to the `mage` build scripts. But it is a change that I believe to be necessary.

### Merging this PR

We need to get this PR matched with a failing wasm-test-suite test case. We then need to show that applying this PR fixes the failing test.

I think I have the failing test in [this branch](https://github.com/owenwaller/vugu/tree/PR/wasm-test-025-valueof) in my fork. Note this test does not call `syscall/js` it calls `vugu/js` only. With the current `vugu/js` that panics so the test fails.

Can you please pull that branch and try it against the copy of `vugu/js` you are proposing in this PR?

You can run things with `mage testSingleWasmTest github.com/vugu/vugu/wasm-test-suite/test-025-valueof`. 

If that test passes with your revised `vugu/js` we can start merging things in.

Dadido3:
> Your [comment](https://github.com/vugu/vugu/pull/332#issuecomment-2620170668) actually agrees with the build tags argument I have been making in Issue #330.

yes, but i never was against the use of build tags.

> You seem to be proposing that this split is enforced through build tags. This is more of less what Issue #330 proposes.

There is a small but important difference, because you suggested to enforce build tags on all `*_gen.go` files. [My proposal was to make them at most optional](https://github.com/vugu/vugu/issues/330#issuecomment-2613684721).

> The proposed file split could also be the higher level of abstraction than a wrapper over a `syscall` package, i.e. `vugu/js` that I proposed [here](https://github.com/vugu/vugu/pull/332#issuecomment-2619829520)
> 
> > it is highly unusual to wrap a syscall package. You would work at a higher level of abstraction and then at build time pull in the required low level package directly

Sure, but this means that someone needs to put work into designing that. And people then have to learn another abstraction instead of using the familiar `syscall/js` API. Also, isn't the maintenance burden even higher than just having a wrapper/alias package?

> Because of the way `vugu` names files, and the significance of the file name structure, I suspect (but have not tried) that file names like this `component_js_wasm.go` won't be matched to a corresponding `component.go` or `component.vugu` file. I think to support this we would have to change `vugugen`. I have not investigated such a change.
> 
> So, for the minute we might need to use the build tags rather than the equivalent file extensions.
> 
> I have checked that you can have a file set named like this `component.vugu`, `component.go`, `implComponent.go` and that `vugugen` will still generate a `component_gen.go` provided everything is in the same package. `implComponent.go` was deliberately chosen so that it did not start with the word `component`.

I haven't checked the internals of `vugugen`, but i don't think you need to modify `vugugen`. As long as there are the usual files (`component.vugu`, `component.go`) it will work as before. And i assume it will ignore any additional files. (But as i have said, i haven't checked any details)

The use of the file extensions `_js.go` and `_native.go` was just an example. It was just to show how it would work in principle. It's probably better not to use these suffixes, but that's also a detail that can be worked out later.

> So, it now seems we both agree we need build tags and that the solution works, after clarification of which files need build tags.

I never disagreed with the use of build tags. But again, only if optional inside of generated files.

> If a developer is creating an application with `vugu` needs to make use of `syscall/js` directly then they need to put the appropriate build tag in the `*.go` file that implements that component and provide a native mock implementation of the same component.
> 
> This seems to break my option (c) constraint [in this comment](https://github.com/vugu/vugu/pull/332#issuecomment-2619829520)

If a developer sets the build tags correctly and provides a mock implementation neither a build nor a test should fail. How does that break your option c?

> a) What does `vugugen` do, and should there be build tags in the `*_gen.go` files?

As i said before, either there should be no build tags, or they should be optional. That's in my opinion the only two options which are feasible. In regards to file naming there probably is no need to touch `vugugen`.

> b) What should the internal `vugu` tests do?

For now, the native part of the test should be put into its own package. The alternative solution would be more build tags, but that's ugly, especially as the native test and the browser side test don't interact with each other.

> c) What do we need to get this PR merged?

Get the test finished up.

> The only reason that `vugu/js` is included as a blank variable of type `js.Value` is to support embedding Go code inside `<script type="application/x-go">` tags in a `*.vugu` file.

As far as i am concerned, it's not important for me to have any `syscall/js` inside the generated files. But that's another discussion.

> _Personal note:_ I don't like this and can think of ways to eliminate it (along with `<script type="application/x-go">` tags. Go code should be in a `*.go` file. But this is supported ATM.

I second that, but for now the script tag is at least needed if you want to import components from other packages. There are probably nicer ways to define additional imports in `component.vugu` files. But that's also another discussion.

> To do this, it needs a change in the file structure of the `vugu` project, and changes to the `mage` build scripts. But it is a change that I believe to be necessary.

Yes, but that's a relatively easy change.

> Can you please pull that branch and try it against the copy of `vugu/js` you are proposing in this PR?

I did, 7 days ago, and just now to be sure:

Fails with the current version of `vugu/js`:

```terminal
...
expected: "Ok"
actual  : "ValueOf: invalid value"
...
FAIL: Test001Simple (0.43s)
```

And it passes with `vugu/js` from this PR:

```terminal
...
PASS: Test001Simple/t1 (0.00s)
...
```

(It's still named `Test001Simple` btw.)

owenwaller:
@Dadido3 

Please see PR #336. This is the full PR that includes the test code that this PR lacks. PR #336 is also based on the current `HEAD` of the `master` branch.

Thank you for confirming that the the test case in [this branch](https://github.com/owenwaller/vugu/tree/PR/wasm-test-025-valueof) passes. It was clear from your first comment that the test cases from `syscall/js` passed but not this test.

If you believe PR #336 needs any change can you please comment in PR #336? Thanks

For the minute I am going to leave this PR open. If PR #336 is the solution then PR #336 is the version that will be merged, and then both PR #332 and PR #336 will be closed.

owenwaller:
@Dadido3 

In answer to your previous question...

> >  At this point I am considering ripping every reference to vugu/js out, and replacing that with syscall/js and then rebuilding everything to see what breaks (I suspect this will break some of the unit tests as well as the static renderer package as a minimum).

> I would be interested to see the result of this.

I have now tried removing all the `vugu/js` references that apply to `GOOS=js GOARCH=wasm` and replacing them with direct calls to `syscall/js`. In the cases where code is compiling as native i.e. tests including `vugu` unit tests these cases still call `vugu/js`. 

All the unit and wasm tests still pass and the changes are largely trivial. It is mostly a matter of file renaming (to add `_js_wasm`) into the filenames for code that is `GOOS=js GOARCH=wasm` specific, and adding `//go:build !js || !wasm` build tags into code that applies to all other platforms, and then separating out any common code - which has no build tags.

In principle `vugu/js` could now be made a internal package that only provides the nonjs implementation. I have not made this change and am not currently proposing it.

I want to thank you for your comments in the discussion above (and in Issue #330). It was your suggestion of a file split that was my "Ah ha" moment. I had completely forgotten that platform specific code can be pulled out into a specially named `*.go` file.

FWIW the change tweaks the `vugugen` code (or at least the generator part) to emit two `*_gen.go` files. One is platform specific to `GOOS=js GOARCH=wasm` the other is for all other platforms. This tweak is particularly crude, but was the simplest and least change way to achieving this outcome.

It is now clear from looking at this code that a substantial amount of refactoring is possible and that this refactoring needs to take place.

So what I am going to propose is this:

a) That we accept PR #336 (the revised change with the test code) simply because my removal change depends on your updated `js/impl-nonjs.go` file. We reject and close PR #332 (as the important changes are in PR #336)
b) I write a full proposal for the removal of `vugu/js` on `GOOS=js GOARCH=wasm` platforms.
c) I create the full PR that demonstrates the removal.

I will now tidy up the change. It really is too crude to publish. I will let you know when it is available.

Dadido3:
> Please see PR #336. This is the full PR that includes the test code that this PR lacks. PR #336 is also based on the current `HEAD` of the `master` branch.
> 
> Thank you for confirming that the the test case in [this branch](https://github.com/owenwaller/vugu/tree/PR/wasm-test-025-valueof) passes. It was clear from your first comment that the test cases from `syscall/js` passed but not this test.
> 
> If you believe PR #336 needs any change can you please comment in PR #336? Thanks
> 
> For the minute I am going to leave this PR open. If PR #336 is the solution then PR #336 is the version that will be merged, and then both PR #332 and PR #336 will be closed.

#336 Looks good to me. The only request i have is to restore the authorship of ebb52c9e2fc39ed197b92c301ad658b82613c1a8. If that's not easily doable, i can create a new commit without `js/impl-js_test.go` that you can cherry-pick.

> FWIW the change tweaks the `vugugen` code (or at least the generator part) to emit two `*_gen.go` files. One is platform specific to `GOOS=js GOARCH=wasm` the other is for all other platforms. This tweak is particularly crude, but was the simplest and least change way to achieving this outcome.

This may also be a workable solution. So basically generate 2 `*_gen.go` files that either import a `syscall/js` mock library on native platforms, or the original `syscall/js` on js platforms? Even though that really means that there will be twice as much generated code. But i suppose you are working towards removing `syscall/js` from the generated file entirely, as you said before.

> a) That we accept PR #336 (the revised change with the test code) simply because my removal change depends on your updated `js/impl-nonjs.go` file. We reject and close PR #332 (as the important changes are in PR #336)
> b) I write a full proposal for the removal of `vugu/js` on `GOOS=js GOARCH=wasm` platforms.
> c) I create the full PR that demonstrates the removal.

I second that. And i'm not against such future solution as long as it doesn't reduce vugu's feature set or causes too much trouble on the user side.

owenwaller:
> https://github.com/vugu/vugu/pull/336 Looks good to me. The only request i have is to restore the authorship of https://github.com/vugu/vugu/commit/ebb52c9e2fc39ed197b92c301ad658b82613c1a8. If that's not easily doable, i can create a new commit without js/impl-js_test.go that you can cherry-pick.

@Dadido3 my apologies it was never my intention to remove you as a commit author. It's an unintended consequence of how the merge was performed.

Your branch is way behind master doesn't include c822d35 in particular - which is a large dependency update. Merging in your change would have undone this change and the commits b8c5882 which is a minor change. Neither of which I want. 

For some reason (which I don't remember now) a `cherry-pick` of ebb529c failed or wasn't going to be easy I can't remember which. So it was then easier to copy your `vugu/js` over what I had locally.

Anyhow we can try to fix this for you. 

Can you please sync to the current `HEAD` of `master` [b8c5882](https://github.com/vugu/vugu/commit/b8c58828a867c06c439027be3c5f6c75e6a9c36d)

You then need to `cherry-pick` 84a9d72 (the test case) which is my test case from [my fork](https://github.com/owenwaller/vugu/commit/84a9d727852173003d4f975214754a94ad1bf544). A `mage testWasm` should fail at this point.

And then `cherry-pick` your commit on top ebb529c. A `mage all` should pass at that point. This will build and test `vugu` (so use the`js/impl-nonjs.go` file and then run the `wasm-test-suite` cases which will use the `js/impl-js.go` file.

Then add a commit that removes the `js/impl-js_tast.go` file as we don't need that.

That I think will leave us with a commit order that shows the failing test being committed before the solution is committed.

Can you then push that as a new PR, and add a note to this PR? I can then add the minor tidy-up of `test-025-valueof` on top of the revised PR.

I'll make sure my changes that remove `vugu/js` can then be `cherry-picked` on to the top of you new PR (or really a new mater with you PR in it).

bradleypeabody:
As soon as the people on this thread are fine with it, I'm good with merging this in as long as we're only fixing vugu/js so it works with recent Go changes (which this PR in it's current state does appear to achieve that).

Dadido3:
I'm fine with either #332 or #337 being merged (The tests in the latter PR still need to be tidied-up/fixed though, which @owenwaller wanted to do).

TBH. I thought the process of getting this merged was nothing more like "a small trip to the grocery store to get a missing ingredient". But instead it feels like putting together a full set of furniture at IKEA, including the trip to the warehouse to pick up all the parts.

owenwaller:
I've just closed my PR#336 as its superseded by PR337, thats the PR we need to merge

I am trying to minimize the number of PR's @bradleypeabody needs to merge in.  

Can I close this issue?

Dadido3:
Yeah, no problem.

PR #336, opened 2025-01-30
Merge PR/332 with the test code

owenwaller:
This commit brings together the changes proposed in PR332 with the test case for the problem from my fork.

See:
Issue #332
[My branch with the test code](https://github.com/owenwaller/vugu/tree/PR/wasm-test-025-valueof)

The commit shows that the changes proposed in PR332 now allow the test
case to pass. The call to `vugu.js.ValueOf(...)` made by the test case no
longer results in a panic.

This commit drops the `js/impl-js_test.go` file which was a copy taken
from the standard library. These test cases rely on using Node.js or similar
to run theses test cases in a wasm runtime. Keeping these tests would
impose another external dependency on the vugu project which we
are keen to avoid.
    
The initial comment in PR332 (https://github.com/vugu/vugu/pull/332#issue-2810677048) makes it clear that the revised `vugu/js` package passes all of the `syscall/js` tests.
    
The test case itself is now correctly compared to the original version in the
branch in my fork.


owenwaller:
Closed as PR #PR 337 is the replacement.

PR #359, opened 2025-04-15
Disable linting on the Github Action (golangci-lint v2.02)

owenwaller:
The now current version of golangci-lint is now causing a large number of errcheck and staticcheck tool failures.

This wasn't the case previously so something has changed in the golangci-lint config.

For the minute linting as been disabled part of the GithubAction target.

The linter erros will need to be cirrected but a large number of them are in packages that we would like to remove e.g. distuil, devutil etc.


owenwaller:
This commit was cherry-picked from https://github.com/owenwaller/vugu/tree/PR/replace-vugu-js-with-syscall-js
Commit #e097057919067dd9bb28bb4cf100f810cc455542

